Issue #333, opened 2014-10-30
Exposing raw Command Execution Time metrics


asibs:
I would like to expose the various Hystrix Metrics to an internal monitoring system, and looked at creating a custom HystrixMetricsPublisher. This would work for most of the information I'm interested in, but I could not find a way to expose the command latencies (both the HystrixCommand.run() execution time and the full end-to-end execution including any queuing time, etc).

At the moment it doesn't seem possible to expose these raw metrics, only rolling aggregates of mean and percentiles. The reason I'm after the raw metrics is because I have a monitoring system which will track metrics over days / weeks, allowing more fine grained historical analysis than could be achieved by exposing already aggregated data.

If this sounds like a useful addition to Hystrix I am happy to fork and implement. Below are some thoughts I had about how this functionality could be implemented, so if this is deemed a suitable & useful addition, any guidance on a preferred approach would be welcome.

I first thought I may be able to achieve this by extending HystrixCommandMetrics with my own implementation, but looking at HystrixCommand and AbstractCommand there appears to be no way to change the HystrixCommandMetrics instance used in AbstractCommand. HystrixCommandMetrics ultimately always gets passed in as null to the AbstractCommand constructor, resulting in a call to HystrixCommandMetrics.getInstance(...) in the metrics initialisation block.

So one approach would be to replace the call to HystrixCommandMetrics.getInstance(...) in AbstractCommand with a call to a factory, and allow others to set the factory they wish to use, thus allowing people to write their own implementations of HystrixCommandMetrics. This approach may not be desirable though, HystrixCommandMetrics is quite an important class in terms of providing HealthCounts to the CircuitBreaker, so allowing others to override the implementation may be deemed risky.

Another approach would be to add one or more methods to HystrixEventNotifier. Something along the lines of a "markCommandDuration(...)" method which would then be called from HystrixCommandMetrics.addCommandExecutionTime(...) and HystrixCommandMetrics.addUserThreadExecutionTime(...), allowing a custom HystrixEventNotifier to also see these execution time events (including the time taken) and pass them onto some external monitoring utility. This approach would make use of the existing ability to plugin a custom HystrixEventNotifier.

mattrjacobs:
[...]

PR #360, opened 2014-12-15
Metrics expire after inactivity in milliseconds

nsowen:
I have quite a number of commands which get created dynamically, therefore the dashboards gets kind of crowded after a few days and we end up calling Hystrix.reset().

Therefore I think it's worth being able to define a "last touch" timestamp within each HystrixCommandMetrics instance and a new optional configuration option for each Command that defines a expire-timeout on per-command basis. Proposed option for expiring metrics automatically after one day:

hystrix.command.default.metrics.expireAfterInactivityInMilliseconds=86400000

Same option for a specific command:

hystrix.command.[CommandKey].metrics.expireAfterInactivityInMilliseconds=86400000

Maybe it is worth discussing the additional CPU cycles required for touch() and expire() in class  HystricCommandMetrics.

Another caveat: existing AJAX-connections with HystrixMetricsStreamServlet will not be updated automatically because its internal LinkedBlockedQueue used holds up to 1000 JSON entries, so also those expired metrics.


mattrjacobs:
This seems like a reasonable request for an environment with a dynamic set of commands.  I haven't encountered that case personally.  I think your point about possible performance degradation is good - calling the `touch()` method on every transition through each command's lifecycle could certainly impact performance.  For that reason, I'd like to work on getting a performance baseline set up with #578 first to actually answer that question.

Also, if you could give me some context around why a dynamic set of commands are necessary for you, that would help me understand your use case better.

nsowen:
Thanks for looking into this. We have a large customer-facing portal with around 1000 req/s, that actually display content from various content providers in realtime. Each user request is multiplexed into several content provider requests.

Each content service will be mapped to (shared) Hystrix pools and each request type (either "page" or "image") will be mapped to a unique Hystrix command, e.g. GetWeatherPage or GetWeatherImage. That way, we will cut off slow responding services automatically without impacting the overall portal performance. The downside is, like you already assumed, a large number of commands and it would be helpful to have the metrics expire automatically.

mattrjacobs:
I now have a performance benchmark merged into master (#750).  Using that, I'm able to get the performance numbers on your change.  Formatting into a Github comment is a bit of a pain, but the overall behavior was a 2-10% throughput decrease when the circuit is closed and a 10-20% throughput decrease when the circuit is open.  

Another approach I thought of while combing through open tickets is the idea of a pluggable metrics strategy.  This was proposed in #333, and seems like a good idea.  Applying it to your use case, you could move the boolean flag for whether you want expiration or not from the command-level to the global level and then choose the proper metrics strategy (one with expiration on all metrics or one without expiration) by registering the appropriate metrics generation plugin.  

Any thoughts on that approach?  I think it has value because, from my perspective, a dynamic command space is not common.  Imposing the performance cost on all consumers seems suboptimal.

mattrjacobs:
@nsowen I'm closing this PR.  If you want to pursue it again, I just merged in a change which allows for a more flexible metrics strategy, so we can work through that if you're interested.

Issue #354, opened 2014-12-10
interruptOnTimeout doesn't work in Hystrix 1.4.0-RC5


aadeon:
When using Hystrix 1.4.0-RC5, the commands are not interrupted when the command times out.
The following naive test shows how to reproduce this:

``` java
public class InterruptTest
{
	private static final Logger LOG = LoggerFactory.getLogger(InterruptTest.class);

	private class SleepingCommand extends HystrixCommand<Void> {

		public SleepingCommand() {
			super(Setter.withGroupKey(HystrixCommandGroupKey.Factory.asKey("mygroup"))
					.andCommandPropertiesDefaults(HystrixCommandProperties.Setter()
							.withExecutionIsolationThreadInterruptOnTimeout(true)
							.withExecutionIsolationThreadTimeoutInMilliseconds(1000)));
		}

		private volatile boolean hasBeenInterrupted;

		public boolean hasBeenInterrupted()
		{
			return hasBeenInterrupted;
		}

		@Override
		protected Void run() throws Exception
		{
			try {
				Thread.sleep(2000);
			}
			catch (InterruptedException e) {
				LOG.error("Interrupted!", e);
				hasBeenInterrupted = true;
			}

			return null;
		}
	}

	@Test
	public void shouldInterruptOnTimeout() throws InterruptedException
	{
		// given
		SleepingCommand cmd = new SleepingCommand();

		// when
		cmd.observe().subscribe();

		// then
		Thread.sleep(3000);
		Assert.assertTrue(cmd.hasBeenInterrupted());
	}
}
```

It works fine with Hystrix 1.3.18.

mattrjacobs:
@aadeon This is a bug, but I think the fix has to be made in conjunction with RxJava.  I talked to @benjchristensen this morning, and there is some active work around thread interruption happening in RxJava at the moment.  

Right now, RxJava is unilaterally deciding to do thread interruption on an unsubscribe.  IIRC, up until 1.02, the choice was to do it, and for 1.0.3+, the choice was not to.  The point is that there's currently no way for Hystrix to specify this decision on a per-instance basis.

As @benjchristensen make progress on this interaction, I'll put notes here.

In order to not hold up RC6, I'm now targeting this to RC7

mattrjacobs:
[...]

PR #379, opened 2014-12-27
Failing unit tests for timeouts on HystrixCommands not causing thread interruptions

mattrjacobs:
This is a PR slightly modified from #354.  Thanks @aadeon for the contribution, I can replicate your failure.

cc @benjchristensen, it seems like we need a way to explicitly control thread interruption of the Observable underlying a HystrixCommand.  Can you take a look and see if there's a workaround in Hystrix, or if corresponding RxJava changes are needed?


mattrjacobs:
Replaced by #593

Issue #290, opened 2014-07-30
Modify ExceptionThreadingUtility to stop adding stack trace as root


mattrjacobs:
See #195 

* Add caller thread stack trace to HystrixRuntimeException
* Override all printStackTrace methods in HystrixRuntimeException to print out the caller thread stack trace

MMalani:
Any idea as to when 1.4.0 will officially be released? Much thanks,,.

mattrjacobs:
This functionality is not going into 1.4.x.  See #451

mattrjacobs:
@MMalani We're looking at the order of a month or so.  At the moment, I've got 1.4.0-RC6 targeted as a bugfix release, 1.4.0-RC7 as an additive feature release, and then will begin production testing at Netflix.  Hopefully, that will go quickly, but if it doesn't, you can expect more issues/pull requests as that gets ironed out.

Issue #289, opened 2014-07-30
Determine performance hit of getting caller thread stack trace on every command execution


mattrjacobs:
See #195

mattrjacobs:
Not worth examining.  Closing.

PR #451, opened 2015-01-07
Removed ExceptionThreadingUtility.  Stack traces are not helpful in an event-loop world

mattrjacobs:
/cc @benjchristensen @KoltonAndrus for review

Issues #289, #290 deal with fixing ExceptionThreadingUtility to correctly attach stack traces (at very high cost).  This doesn't seem worthwhile, especially since, even if we did that properly, they would become useless as soon as Hystrix is used in an event-loop system like Netty.


benjchristensen:
I'm fine with removing it.

benjchristensen:
:+1: on the changes. 

:-1: on non-deterministic tests that I wrote.

PR #453, opened 2015-01-07
Resettable command and thread pool defaults

### LINKED ISSUE #280: Hystrix.reset() doesn't reset default command properties
Hystrix version: 1.4.0-RC4
The documentation of `com.netflix.hystrix.HystrixCommand#HystrixCommand(com.netflix.hystrix.HystrixCommand.Setter)` says: 

"Properties passed in via Setter#andCommandPropertiesDefaults or Setter#andThreadPoolPropertiesDefaults are cached for the given HystrixCommandKey for the life of the JVM or until Hystrix#reset() is called."

Unfortunately calling `Hystrix.reset()` doesn't reset all propery defaults, and , in particular `executionIsolationThreadTimeoutInMilliseconds` set for instance this way:

    protected MyCommand(final int timeout)
	{
       super(Setter.withGroupKey(COMMAND_GROUP).andCommandPropertiesDefaults(HystrixCommandProperties.Setter()
                 .withExecutionIsolationThreadTimeoutInMilliseconds(timeout)));

    }

Because of this either the documention or `Hystrix.reset()` should be corrected. It would be nice, however, if  `Hystrix.reset()` really resets the timeouts. This may be useful in (unit) tests.

mattrjacobs:
Closes #280

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/Hystrix.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/Hystrix.java
@@ -5,6 +5,7 @@ import java.util.NoSuchElementException;
 import java.util.concurrent.TimeUnit;
 
 import com.netflix.hystrix.strategy.HystrixPlugins;
+import com.netflix.hystrix.strategy.properties.HystrixPropertiesFactory;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


mattrjacobs:
Closes #280

PR #456, opened 2015-01-07
Bypass fallback/circuit breaker for BadRequestExceptions from ExecutionHook

gerrit-hntschl:
Hi,

recently introduced Hystrix into an existing codebase and like it a lot. That codebase uses a custom `BusinessException` to signal bad user input etc. We would like to automatically wrap the `BusinessException`s into a `HystrixBadRequestException` to isolate Hystrix-Exception handling and not having to touch all the existing code.
However, turning an `Exception` into a `HystrixBadRequestException` inside `HystrixCommandExecutionHook#onRunError` did not side-step the fallback machinery. This pull request fixes that.

Cheers,
Gerrit


benjchristensen:
On a quick check this looks good. @mattrjacobs can confirm further. 

Thanks @gerrit-hntschl

mattrjacobs:
Yeah, thanks for including the unit test, @gerrit-hntschl!  Merging now...

PR #506, opened 2015-01-14
Backport of #327 onto 1.3.x

mattrjacobs:
This pull request asserts that the proper calls on HystrixCommandExecutionHook are made.  The bugfixes identified by these assertions are:

* HystrixBadRequestException flow was not invoking ```onError()```
* run() throwing an Exception with a successful fallback was invoking ```onComplete()``` twice

PR #508, opened 2015-01-14
Backport of #327

mattrjacobs:
This is an updated version of #506.  It adds assertions on hook execution.


mattrjacobs:
This pull request asserts that the proper calls on HystrixCommandExecutionHook are made.  The bugfixes identified by these assertions are:

* HystrixBadRequestException flow was not invoking ```onError()```
* run() throwing an Exception with a successful fallback was invoking ```onComplete()``` twice

Issue #510, opened 2015-01-14
onRunSuccess/onThreadComplete hooks called in timeout flow


mattrjacobs:
In 1.3.x, under thread isolation, when a HystrixCommand times out, the ```onRunSuccess```/```onThreadComplete``` hooks are invoked by the Hystrix thread eventually, even if the calling thread has already moved through the fallback flow.

In master/1.4, only the ```onThreadComplete``` hook is invoked (eventually).  

/cc @benjchristensen, @KoltonAndrus What should the semantics be?  I'd like to decide that and then enforce it in both 1.3.x and master

benjchristensen:
I think it should call eventually whenever it does complete. A thread always ends (unless it's an infinite loop). 

It is clearer for it to eventually be called than to have cases where it only starts but never "ends". Also I don't want to break 1.3 behavior.

mattrjacobs:
Yeah, that's the way I was leaning, but didn't want to bias your response :)  This will be a change in master only then.  I'll also add some Javadoc to clarify this case, as it's not obvious.

mattrjacobs:
Added Javadoc in #512

PR #511, opened 2015-01-14
Matching hook orders from 1.3.x

mattrjacobs:
Uncovered Issue #510 during this investigation.  Will update this PR once that is resolved.


mattrjacobs:
2 bugs uncovered by copying over hook execution orders from 1.3.x branch:

* Short-circuiting with fallback was not triggering ```onComplete```
* Timeout (with or without fallback) was not triggering ```onRunSuccess``` when the Hystrix thread ran successfully but exceeded the command timeout

Issue #514, opened 2015-01-16
Timeout mechanism doesn't work for command which is queued but not blocked upon


mattrjacobs:
Found this during investigation of #509.  I will create an independent PR for this issue.

The gist of it is that a user may invoke ```HystrixCommand.queue()``` to produce a ```Future```.  If the user calls ```.get()``` on the ```Future```, then the timeout mechanism works properly.  If the user does other work on the calling thread and lets the command complete naturally, the timeout never fires.

From a cursory code review, that seems possible, as all of the timeout setup occurs on the path in which ```.get()``` or ```.get(int, TimeUnit) is invoked

mattrjacobs:
I've got a local branch that removes the blocking timeout and ```performBlockingGetWithTimeout``` path.  It uses the ```HystrixTimer``` nonblocking timeout approach for all timeouts.  This addresses the issue above, but I still need to performance test it.

PR #515, opened 2015-01-16
Unit test demonstrating that a timeout does not fire properly when command.queue() is called with no blocking performed on the resulting Future

mattrjacobs:
Demonstrates issue from #514.  Fix forthcoming...


mattrjacobs:
Replaced by #580

PR #516, opened 2015-01-16
Updated HystrixServoMetricsPublisher initalization of singleton

jewzaam:
Noticed that singleton initialization is not great when debugging an application using this project.  Don't know if it will fix the problem we're seeing, but this initialization is more robust than what was there before.


mattrjacobs:
Thanks for the contribution, @jewzaam!.  Merging.

PR #524, opened 2015-01-19
Change calls from getExecutedCommands() to getAllExecutedCommands()

kevinvandervlist:
The calls to the deprecated getExecutedCommands() from HystrixRequestLog have been updated to getAllExecutedCommands(). This method returns a HystrixInvokableInfo instead of a HystrixCommand, so a few tests and examples have been updated to reflect this changed type as well.


mattrjacobs:
Thanks for the contribution @kevinvandervlist !

Issue #346, opened 2014-11-27
HystrixMetricsStreamServlet.destroy() not invoked until container timeout


codependent:
__Scenario:__
1. JEE App with hystrix + hystrix-metrix-event-stream (v1.3.18). 
2. Hystrix dashboard requesting data to the App.
3. When we stop App in our server we see that the following loop doesn't stop: 

    HystrixMetricsStreamServlet.handleRequest(){
    ...  
       while (poller.isRunning() && !isDestroyed) {

The servlet's destroy() method is not called, thus **isDestroyed is not modified until timeout**. This makes sense since, according to the servlet spec: *void destroy() - Called by [...] This method is only called once all threads within the servlet's service method have exited or after a timeout period has passed.*

The problem is that since there is a running request, whenever we want to republish/stop our application, we have to wait 60 seconds until our container (Websphere Application Server 8) forces the destruction of the servlet.

mattrjacobs:
@jboyd01 I don't have a great way of testing this, as I'm not familiar with Websphere or WAS.  Are you (or anyone else) still interested in submitting a pull request?

mattrjacobs:
Moving this out of RC6.  I'm happy to merge in a PR from someone who experiences this issue, but don't have enough context to work on it myself.

codependent:
@mattrjacobs I could (and will) submit jboyd01's solution as a PR. We are planning to implement Hystrix-enabled services in the next months and will definitely need that workaround.

mattrjacobs:
Great, thanks @codependent!

lasselasse:
We also just hit this issue running the metrics stream servlet from within embedded Jetty.
Very excited to see this being addressed soon!

mattrjacobs:
@codependent / @lasselasse I just merged this into master.  Are you able to test @jboyd01 's solution currently or do you need an official RC release?

codependent:
@mattrjacobs I don't really need a RC thanks. I'll test it tomorrow (out of the office right now) and let you know if it's OK.

mattrjacobs:
Great, thanks @codependent

codependent:
@mattrjacobs I just tested it, and it works OK. Thanks to @jboyd01 for this workaround.

lasselasse:
@mattrjacobs tested the patch and it works great, thanks!

[...]

PR #530, opened 2015-01-20
Add shutdown hook for WebSphere and checkError() to detect disconnected client

jboyd01:
fixes https://github.com/Netflix/Hystrix/issues/346

WebSphere won't shutdown a servlet until after a 60 second timeout if there is an  instance of the servlet executing a request.  Add a shutdown method to enable a hook to notify Hystrix to shutdown.  You must invoke this method at app server shutdown, perhaps from another servlet's destroy() method.

Also added explicit checkError() in poller loop to detect disconnected client.


mattrjacobs:
Thanks @jboyd!

dsukner:
This fix does not seem to have fixed the WebSphere issue.

"Also added explicit checkError() in poller loop to check for
disconnected client."

Issue #509, opened 2015-01-14
onThreadStart/onThreadComplete doesn't always match


mattrjacobs:
When running a snapshot of current master in a Netflix canary, I've found that the number of ```onThreadStart``` hook calls and ```onThreadComplete``` hook calls don't always line up.  I'll be adding unit tests until I figure out the specific execution flow which invokes ```onThreadStart``` but not ```onThreadComplete```

mattrjacobs:
Resolved by #531

PR #531, opened 2015-01-20
Move onThreadStart execution hook after check that wrapping thread timed out

mattrjacobs:


mattrjacobs:
In Netflix production testing, this fix resolved #509

Issue #368, opened 2014-12-17
Travis CI intermittent test failures


mattrjacobs:
com.netflix.hystrix.strategy.HystrixPluginsTest > testPropertiesStrategyViaProperty FAILED
    java.lang.AssertionError at HystrixPluginsTest.java:165

com.netflix.hystrix.strategy.HystrixPluginsTest > testConcurrencyStrategyViaProperty FAILED
    java.lang.AssertionError at HystrixPluginsTest.java:100

com.netflix.hystrix.strategy.HystrixPluginsTest > testMetricsPublisherDefaultImpl FAILED
    java.lang.AssertionError at HystrixPluginsTest.java:119

com.netflix.hystrix.strategy.HystrixPluginsTest > testEventNotifierViaProperty FAILED
    java.lang.AssertionError at HystrixPluginsTest.java:75

com.netflix.hystrix.strategy.HystrixPluginsTest > testConcurrencyStrategyViaRegisterMethod FAILED
    java.lang.IllegalStateException at HystrixPluginsTest.java:94

com.netflix.hystrix.strategy.HystrixPluginsTest > testMetricsPublisherViaRegisterMethod FAILED
    java.lang.IllegalStateException at HystrixPluginsTest.java:124

com.netflix.hystrix.strategy.HystrixPluginsTest > testCommandExecutionHookViaRegisterMethod FAILED
    java.lang.IllegalStateException at HystrixPluginsTest.java:48

com.netflix.hystrix.HystrixObservableCommandTest > testSemaphorePermitsInUse FAILED
    java.lang.AssertionError at HystrixObservableCommandTest.java:2068

com.netflix.hystrix.strategy.HystrixPluginsTest > testRequestContextViaPluginInTimeout FAILED
    java.lang.IllegalStateException at HystrixPluginsTest.java:178

com.netflix.hystrix.HystrixObservableCommandTest > testRejectedViaSemaphoreIsolation FAILED     java.lang.AssertionError at HystrixObservableCommandTest.java:5066

com.netflix.hystrix.HystrixObservableCollapserTest > testTwoRequests FAILED
    java.lang.AssertionError at HystrixObservableCollapserTest.java:78

com.netflix.hystrix.HystrixObservableCommandTest > testExecutionHookThreadTimeoutSuccessfulFallbackRunFailure FAILED
    java.lang.AssertionError at HystrixObservableCommandTest.java:3038

mattrjacobs:
[...]

PR #567, opened 2015-01-25
Commenting out the flaky HystrixTest and HystrixPluginsTest until someone has time to de-flake them

mattrjacobs:


mattrjacobs:
See #368

Issue #373, opened 2014-12-18
Upgrade to Java 7


mattrjacobs:
Java 6 is EOL and this helps address #372

rspieldenner:
The new nebula.netflixoss build system defaults to java 7 if that adds anything to the conversation. We can still set to java 6 or java 8.

mattrjacobs:
Closed by #577

PR #577, opened 2015-01-28
Upgrade to Java7

mattrjacobs:
Addresses #373.  

Is this the right way to perform the upgrade w.r.t Gradle?  /cc @rspieldenner @quidryan


mattrjacobs:
Merging this now, as it passes all tests.  If there's further work around Gradle setup, please open a new issue

rspieldenner:
you don't need the sourceCompatibility and targetCompatibility = 1.7 as those are the defaults but it won't hurt anything

mattrjacobs:
Great, thanks for the confirmation @rspieldenner

PR #596, opened 2015-02-02
Fixed HystrixContextScheduler to conform with RxJava Worker contract

akarnokd:
Fixed issue in #354 by making ```HystrixContextScheduler``` conform with RxJava's Worker contract. See also #593.

To avoid code duplication, it uses the ```rx.internal.schedulers.ScheduledAction```. I don't know how close Hystrix follows RxJava releases, but I'll make steps to have ```ScheduledAction``` part of the stable RxJava API eventually.

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/strategy/concurrency/HystrixContextScheduler.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/strategy/concurrency/HystrixContextScheduler.java
@@ -15,16 +15,12 @@
  */
 package com.netflix.hystrix.strategy.concurrency;
 
-import java.util.concurrent.RejectedExecutionException;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.*;
 
-import rx.Scheduler;
-import rx.Subscription;
+import rx.*;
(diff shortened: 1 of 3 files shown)


mattrjacobs:
This is great @akarnokd , thanks for the contribution!  

The other piece of this issue is that a HystrixCommand has configuration to optionally perform a thread interruption on some ```unsubscribe```s.   From what I can tell ```SchedulerAction.FutureCompleter``` already has a unsubscribe policy where thread interrupts happen if and only if the thread is different from the current.  The policy I'm describing for Hystrix would be different from that, and determined at schedule-time.

I'm sure I could whip up a different hack for that - but is there a better way to make the thread-interruption policy more configurable?

akarnokd:
I'm currently proposing a change for ```ScheduledAction``` in https://github.com/ReactiveX/RxJava/pull/2579 which adds a system-property to enable/disable interruption globally. I'll add some interrupt-support there.

mattrjacobs:
I'm merging this in.  I'll open up a separate issue for the configurability of the thread-interrupt behavior.  Then, depending on how RxJava and Hystrix release cycles line up, I'll likely just use the new behavior in RxJava (https://github.com/ReactiveX/RxJava/pull/2579/files).

Issue #302, opened 2014-09-01
1.4: RequestLog with streaming results


benjchristensen:
Now that we support an Observable we can have a stream of results. The RequestLog currently will mark SUCCESS for each onNext rather than the terminal event. 

Everything behaves fine for scalar responses, just need to fix it for a use case where the Observables returns more than 1 value.

mattrjacobs:
The plan here is to add 2 new states to the HystrixRequestLog: ```EMIT``` and ```FALLBACK_EMIT```.  These represent the ```construct()``` and ```resumeWithFallback()``` methods invoking ```onNext()```.

Some possible new RequestLog strings:
[EMITx10, SUCCESS]
[EMITx3, FAILURE]
[EMITx4, FAILURE, FALLBACK_EMITx3, FALLBACK_SUCCESS]
[EMITx1, FAILURE, FALLBACK_EIMTx10, FALLBACK_FAILURE]

mattrjacobs:
One question which I'm not sure about is if we want these new event types in the RequestLog of a ```HystrixCommand```.  It doesn't give any info, as we will always see an ```onNext```/```onCompleted``` pair (```EMIT```/```SUCCESS```) together.

But it also feels a little strange to special-case this.

My initial inclination is to only present these events for ```HystrixObservableCommand```, both to preserve the RequestLog as-is for this case, and because it's not especially meaningful

benjchristensen:
I agree it should not be on the existing HystrixCommand, only ObservableCommand.

mattrjacobs:
To make this concrete, here are some sample request logs for HystrixObservableCommand.  HystrixCommand request logs are unchanged, as described above.

* ```TestCommandWithMultipleValues[EMITx3, SUCCESS][30ms]```
* ```TestPartialSuccess[EMITx3, FAILURE][39ms]```
* ```TestPartialSuccessWithFallback[EMITx3, FAILURE, FALLBACK_EMITx4, FALLBACK_SUCCESS][39ms]```

mattrjacobs:
Done in #604

PR #604, opened 2015-02-04
Added 2 new HystrixEventTypes: EMIT and FALLBACK_EMIT

mattrjacobs:
** These events only get generated in the HystrixObservable case.
** They correspond to an onNext() by construct() and onNext() by resumeWithFallback(), respectively


mattrjacobs:
Addresses #302

PR #643, opened 2015-02-06
Stress-test HystrixObservableCollapser

mattrjacobs:


mattrjacobs:
Didn't see a problem with test that failed as part of #368

Issue #598, opened 2015-02-02
Make thread-interruption of Hystrix thread depend on property value


mattrjacobs:
At the moment, thread-interrupt is controlled by the internals of the Rx Scheduler infrastructure.  @akarnokd has got an addition to the API in https://github.com/ReactiveX/RxJava/pull/2579/files that will expose this functionality for the ```HystrixContextScheduler``` to use.  Once the next RxJava release comes out, move to that API.

Also, see #593, #596 for history

mattrjacobs:
Fixed in #647

PR #646, opened 2015-02-06
Tie command property to thread interrupt

mattrjacobs:
Partially done - addresses #354, #598


mattrjacobs:
Declined - don't like this approach.  Talked with @benjchristensen to find better option

PR #647, opened 2015-02-07
Tie command property to thread interrupt

mattrjacobs:
Addresses #354 and #598.  I think this causes some Javanica unit tests to fail.  Will address those next


mattrjacobs:
There's also an open issue about how to represent thread-interruption in ScheduledActions in RxJava.  Relates to ReactiveX/RxJava#2579

mattrjacobs:
I looked for a few minutes at the failing unit tests in hystrix-javanica and didn't see an obvious reason why they are failing now.

I could use some help from those more familiar with hystrix-javanica.  Any ideas, @dmgcodevil, @sawano / others?

mattrjacobs:
Merging this so I can make forward progress on top of this.  Opened up a new issue (#663) to track the Javanica unit tests

Issue #300, opened 2014-09-01
1.4: executionTimeout property for async


benjchristensen:
We need a property on the semaphore isolated path for timeout now that we support async timeout.

Currently in the 1.4 RCs it uses the threadIsolated property for both, but that's confusing to set it for semaphoreIsolated configs.

mattrjacobs:
@KoltonAndrus I don't see this property referenced anywhere in the execution flow.  Are you able to pick this back up and get that wired in?

mattrjacobs:
After further thought, I think the ideal design is to have 1 timeout value defined, regardless of whether the command is sempahore- or thread-isolated.  

Since only the async timeout mechanism is used now (as of #579), it's straightforward to wire a signal timeout into the AbstractCommand flow.  The only interesting thing is the configuration API.  My view is that the config setting should be agnostic of semaphore or thread isolation, and the config that is thread-specific (```HystrixCommandProperties.executionIsolationThreadTimeoutInMilliseconds```) should be deprecated.

mattrjacobs:
Closed by #664

PR #664, opened 2015-02-10
Favor timeout property that is not thread-specific. (execution.timeoutInMilliseconds)

mattrjacobs:
Also deprecate usages of thread-specific setting (like execution.thread.isolation.timeoutInMilliseconds).

This addresses #300


holy12345:
Hi @mattrjacobs 
I have a question want your confirmation.
In the PR you submitted I saw an addition of an  `execution.timeoutInMilliseconds`  property, which represents the timeout period (regardless of whether the isolation strategy is semaphore or thread).
I understand that this change can be seen in `1.4RC` or later.

But my current version is 1.5.12. The `execution.timeoutInMilliseconds` property does not exist.

I saw the following code in `HystrixCommandProperties`

```
 //this property name is now misleading. //TODO figure out a good way to deprecate this property name
this.executionTimeoutInMilliseconds = getProperty(propertyPrefix, key, "execution.isolation.thread.timeoutInMilliseconds", builder.getExecutionIsolationThreadTimeoutInMilliseconds(), default_executionTimeoutInMilliseconds);
```

By javadoc and my local validation I got the following conclusion:

- In the `1.5.12` version we did not `execute.timeoutInMilliseconds` this parameter.

- Currently we are still using execution.isolation.thread.timeoutInMilliseconds to represent the timeout (regardless of whether the isolation strategy is semaphore or thread, as javadoc said, the name of this parameter is confusing)

Am i right? Looking forward to your reply. Thanks again, best wishes

Issue #54, opened 2012-12-15
Rolling Metric for Max Concurrent Requests


benjchristensen:
We already have this for thread executions but it would be good to have for the command in general (thread or semaphore isolated).

The value of this is in tracking down bursts. Even though we are generally tracking metrics at the second or subsecond level the concurrent bursts at the millisecond level can be missed and make it hard to determine where rejections are coming from.

mattrjacobs:
Addressed by #667

PR #667, opened 2015-02-10
Added a rolling max counter for command execution

mattrjacobs:
Addresses #54


LuboVarga:
sidenote: HystrixCodaHaleMetricsPublisher is not exporting this metrics probably.

PR #693, opened 2015-02-23
This fixes #692. If ratePerSecond is NaN, set errorThenVolume to -1

### LINKED ISSUE #692: Hystrix Dashboard sorts NaN.0/s between error and high-volume when sorting on "error then volume"
When sorting on "error then volume" and ratePerSecond is "NaN", then "errorThenVolume" will become 0.NaN and be sorted right after the monitor-items with errors, but before monitor-items with volume. 

This can be fixed by checking if ratePerSecond is NaN, and then set errorThenVolume to a negative value, or write a custom sorting for tsort().

nilsmagnus:


mattrjacobs:
Thanks for the contribution, @nilsmagnus !

PR #700, opened 2015-03-02
Example code for publishing to Graphite

jboyd01:
Time saver for those wishing to test out easy Graphite integration vs digging through Servo doc and example code


mattrjacobs:
Thanks @jboyd01

PR #711, opened 2015-03-10
Use Archaius for Hystrix plugin setup

### LINKED ISSUE #92: Is there a way to automatically pass thread static information to the Hystrix Command thread pool?
Hello,

I have a number of 3 REST-based micro services. Each of these belongs to a logical layer. A Web app calls the service on the upper layer and the call goes down to the second layer and finally the third one where data are obtained and get responded all the way back. I need to log a common RequestId (as a GUID) that identifies a single web request in all layers. This request id is created on the first service and as a Header it automatically passes from each http client to the next service and from the service to the next client using Client & Servlet filters and a Logback MDC instance. So, in each layer, if the Servlet filter does not get a RequestId header, it creates it, and it puts it in the MDC for the current thread. Whenever an http client request is made, the Client Filter, reads the request id from the MDC, creates a header and passes it to the request and the chain goes on.

If i use a Hystrix Command as a wrapper to the actual http client call, the chain breaks. Apparently, the threads created by a Hystrix Command are not "children" of the thread that called its constructor and therefore the MDC values do not propagate.

Is there a solution to this problem? I read about the execution hooks. Can they offer a solution?

More info at http://logback.qos.ch/manual/mdc.html#autoMDC and http://logback.qos.ch/manual/mdc.html#managedThreads

eirslett:
> **Use Archaius for Hystrix plugin setup**
>
> Currently, Hystrix lets you define custom strategies in two ways:
> 1) with System properties
> 2) with bootstrapping via HystrixPlugins.getInstance().registerXXX
> If neither is specified, the default strategy is used.
> 
> This change replaces hardwired System.getProperty calls with lookup
> via Archaius. So one can override the plugin strategies used with
> any configuration provider Archaius is configured with.
> 
> Most importantly, it has the effect of checking any file called
> "config.properties" on the classpath as well as using system properties.
> This lets you configure plugin strategies without having to write
> Java code for it, or having to run the application with additional
> system properties.
> 
> Fixes #92

According to https://github.com/Netflix/Hystrix/issues/92#issuecomment-76773888, Hystrix is supposed to load its configuration from Archaius, but apparently that is not the case with HystrixPlugins. This PR fixes that.

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/strategy/HystrixPlugins.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/strategy/HystrixPlugins.java
@@ -15,8 +15,10 @@
  */
 package com.netflix.hystrix.strategy;
 
+import java.io.IOException;
 import java.util.concurrent.atomic.AtomicReference;
 
+import com.netflix.config.ConfigurationManager;
 import com.netflix.config.DynamicPropertyFactory;
 import com.netflix.hystrix.strategy.concurrency.HystrixConcurrencyStrategy;
 import com.netflix.hystrix.strategy.concurrency.HystrixConcurrencyStrategyDefault;
@@ -50,7 +52,12 @@ public class HystrixPlugins {
     /* package */ final AtomicReference<HystrixCommandExecutionHook> commandExecutionHook = new AtomicReference<>();
 
     private HystrixPlugins() {
-
+        try {
+            // Load configuration from hystrix-plugins.properties, if that file exists
+            ConfigurationManager.loadCascadedPropertiesFromResources("hystrix-plugins");
+        } catch (IOException e) {
+            // fail silently
+        }
     }
 
     public static HystrixPlugins getInstance() {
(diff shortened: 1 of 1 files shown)


eirslett:
Ideally, it would also check for "hystrix.properties", and use that if found on the classpath?

eirslett:
@michaelsembwever @chriswk @hennings would this help us integrate Zipkin and Hystrix without boilerplate in the userland code?

mattrjacobs:
Thanks for the contribution, @eirslett .  This looks like a good change, but I need to play around with it and understand how I would use it before I merge.  Along those lines, if you could include some sample setup on how you intend to use it, that would probably be helpful (for me and for the docs)

eirslett:
Here's a simple demo library, 

https://github.com/eirslett/pull-request-illustration

The idea is that you would put this library in your classpath, and it would bootstrap everything for you.
 (The alternative would be to set up the ConcurrencyStrategy manually in Java land, but that would be more intrusive, and violate Zipkin's "It Should Just Work" philosophy)
Does that clear things up?

mattrjacobs:
This looks like a great change.  Thanks for the patience, @eirslett .  I'll add some documentation around this to the Wiki shortly.

mattrjacobs:
@eirslett I just added some documentation around this change, as I'm about to release it in 1.4.3.  For now, I've got a pointer to your illustrative repo - thanks very much for that!  At some point, I will add some examples of this to the Hystrix repo.

Can you take a look and see if it makes sense to you? : https://github.com/Netflix/Hystrix/wiki/Plugins#how-to-use

/cc @DavidMGross

eirslett:
Sure, looks good to me!

PR #718, opened 2015-03-24
Add a new interface, `DefaultRequestLogStrategy`, to disable requestLog for the concurrency strategy.

axiak:
In https://github.com/Netflix/Hystrix/commit/706a0f267ed65a215de1d813598c6c72b00f76eb, the removal of checking of the request log is enabled before the request log initialization has made it impossible to override the concurrency strategy without wrapping all creation/execution of hystrix commands in hystrix context initializations.

This change adds an opt-in interface to use the default request log behavior, instead of forcing the implementor of a concurrency strategy to also change the request log behavior. The advantage of a separate interface is that this is class-file backwards compatible - already compiled concurrency strategies will still work with this commit and will retain the current behavior.

@mattrjacobs

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/AbstractCommand.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/AbstractCommand.java
@@ -27,6 +27,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 
+import com.netflix.hystrix.strategy.concurrency.DefaultRequestLogStrategy;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -258,7 +259,7 @@ import com.netflix.hystrix.util.HystrixTimer.TimerListener;
         this.requestCache = HystrixRequestCache.getInstance(this.commandKey, this.concurrencyStrategy);
 
         /* store reference to request log regardless of which thread later hits it */
-        if (concurrencyStrategy instanceof HystrixConcurrencyStrategyDefault) {
+        if (concurrencyStrategy instanceof DefaultRequestLogStrategy) {
             // if we're using the default we support only optionally using a request context
             if (HystrixRequestContext.isCurrentThreadInitialized()) {
                 currentRequestLog = HystrixRequestLog.getCurrentRequest(concurrencyStrategy);
--- a/hystrix-core/src/main/java/com/netflix/hystrix/strategy/concurrency/DefaultRequestLogStrategy.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/strategy/concurrency/DefaultRequestLogStrategy.java
@@ -1,0 +1,8 @@
+package com.netflix.hystrix.strategy.concurrency;
+
+/**
+ * Indicates that the HystrixConcurrencyStrategy does not use a custom request log,
+ * therefore threads do not need to initialize the request log.
+ */
+public interface DefaultRequestLogStrategy {
+}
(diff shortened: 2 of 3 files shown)


axiak:
(This build failure seems unrelated to this PR: https://netflixoss.ci.cloudbees.com/job/NetflixOSS/job/Hystrix/job/Hystrix-pull-requests/64/testReport/junit/com.netflix.hystrix.contrib.rxnetty.metricsstream/HystrixMetricsStreamHandlerTest/initializationError/)

benjchristensen:
If backwards compatibility was broken as this seems to say then we need to fix that.

axiak:
Hey @benjchristensen it's not clear to me how the change would have affected https://github.com/Netflix/Hystrix/blob/master/hystrix-contrib/hystrix-rx-netty-metrics-stream/src/test/java/com/netflix/hystrix/HystrixCommandMetricsSamples.java#L53 ?

axiak:
Looking around it seems "Bad <init> method call from inside of a branch" is a problem with java8's verification

axiak:
@benjchristensen to prove that the test failure was a change rather than my PR, I created a dummy PR which just added a newline to a readme file (https://github.com/Netflix/Hystrix/pull/721). Same error, in the same location. Is it perhaps an upgrade to jdk 7u65? (https://bugs.openjdk.java.net/browse/JDK-8051012)

(Having jenkins display the java version would be helpful in debugging these things.)

mattrjacobs:
@axiak - Thanks for the contribution!

The build failures both look unrelated and due to troubles on the CI hosts.  I think I got Cloudbees fixed up (see #721) for details, but Travis is still not in a good state at the moment.

mattrjacobs:
To make sure I understand this change, I'll restate it.  Please correct anything I get wrong.

In 1.3.x, there was no caching per-command of the HystrixRequestLog in the HystrixCommand, instead it got looked up from the HystrixRequestVariableHolder each time it was referenced.  Furthermore, this reference was wrapped in a check to see if the requestLog was enabled.

In 1.4.x, upon command construction, the `currentRequestLog` value is computed and cached so that this lookup only happens once over a command's lifetime.

In both 1.3.x and 1.4.x, there is a branch on the type of concurrency strategy.  

The first case is that the concurrency strategy is the default concurrency strategy.  If this is true, then a check is done to see if a `HystrixRequestContext` is being used.  If so, then the `currentRequestLog` is looked up/created and cached.  If not, then it's set to null (as a request log can't be used without a request context)

The second case is that a custom concurrency strategy is provided.  In this case, then a lookup is done in the request cache for the request log.  This assumes that a `HystrixRequestContext` has been setup, even if a request log is not desired.

So in order for command construction not to fail, it's necessary for that concurrency strategy to perform the setup/teardown of the request context.

Did I get that right?

mattrjacobs:
In light of the above, your change is completely valid.  It does force a concurrency provider to know about and use the DefaultRequestLogStrategy.

Another change that seems like it should work is to wrap all of the logic which sets the `currentRequestLog` in a check on the `properties.requestLogEnabled().get()` value.  If that was set to false, then the logic never gets triggered, and the lack of a `HystrixRequestContext` shouldn't break anything.

Does that solution sound workable to you?
If so, do you still prefer the DefaultRequestLogStrategy approach?

axiak:
@mattrjacobs you have that right. I'm happy with either reverting the previous change to guard the entire request log initialization with the property lookup or this change. The reason for the strategy in this PR was that it's completely backwards compatible. By opting for a new interface over a method it shouldn't even require a compilation of dependencies built on prior versions of hystrix.

I guess I assume that the initialization in the creation of a command removed its guard on purpose.. maybe something in light of the observer based commands? I'm not entirely sure how the HystrixRequestContext works in the observer framework if the actions could occur in different threads, but maybe that's why it forces it to initialize so that the command will always have a reference to the request log thread.

mattrjacobs:
Your point on backwards-compatibility is a good one.  However, I have a preference for adding the check back, because it places less burden on a plugin writer.  

My guess is this was an unintentional change, and hadn't affected any concurrency plugins that we've used in practice, so we didn't spot the regression.  It also points out that I should add some unit tests around concurrency plugins without a request-context.  I'm guessing there aren't any of those, or else this regression would have been caught. 

Can you resubmit with the requestLogEnabled check?  I'll merge that and release 1.4.2.

axiak:
Superseded by https://github.com/Netflix/Hystrix/pull/724

PR #721, opened 2015-03-25
dummy change

axiak:


mattrjacobs:
There was a change to the Cloudbees config that updated the JDK.  I reverted that change and got that passing.  Looking at Travis now...

mattrjacobs:
In the Travis case, it appears that the http://jcenter.bintray.com/net/sf/ezmorph/ezmorph/1.0.6/ezmorph-1.0.6.pom URL is returning a 500.  Not sure why that is.  If it persists tomorrow, will try to escalate at bintray.

PR #723, opened 2015-03-25
fixed AopUtils#getDeclaredMethod to look for method also in superclasses

vitalatron:
The HystrixCommandAspect fails with NPE if the method annotated with @HystrixCommand is implemented in a superclass. Since the Class#getDeclaredMethod can't see inherited methods but in spite of Class#getMethod can see private ones, I performed a recursive search up by class hierarchy.


vitalatron:
Looks like build fails for the same reason as in https://github.com/Netflix/Hystrix/pull/718

PR #724, opened 2015-03-25
Check if the request log is enabled before requiring the context is initialized.

axiak:
In https://github.com/Netflix/Hystrix/commit/706a0f267ed65a215de1d813598c6c72b00f76eb,
the guard to initialize the request log based on the property was removed. This
causes a dependency on hystrix context initialization for all custom concurrency
strategies.

Here we add that back in, along with a test to ensure that a custom strategy could
still run without an initialized context.

This PR supercedes https://github.com/Netflix/Hystrix/pull/718/

/cc @mattrjacobs

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/AbstractCommand.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/AbstractCommand.java
@@ -257,21 +257,25 @@ import com.netflix.hystrix.util.HystrixTimer.TimerListener;
         /* setup the request cache for this instance */
         this.requestCache = HystrixRequestCache.getInstance(this.commandKey, this.concurrencyStrategy);
 
-        /* store reference to request log regardless of which thread later hits it */
-        if (concurrencyStrategy instanceof HystrixConcurrencyStrategyDefault) {
-            // if we're using the default we support only optionally using a request context
-            if (HystrixRequestContext.isCurrentThreadInitialized()) {
+        if (properties.requestLogEnabled().get()) {
+            /* store reference to request log regardless of which thread later hits it */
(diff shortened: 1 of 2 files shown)


mattrjacobs:
Looks great - thanks for adding the unit test as well that exercises this case.  I've gotten Cloudbees happy, so I'm merging this and will release it ASAP as 1.4.2

axiak:
Thanks!

axiak:
Hey @mattrjacobs, how long do you think it'll be until 1.4.2 is published? I want to upgrade all of our stuff ASAP. Thanks again

mattrjacobs:
This was released yesterday afternoon: https://github.com/Netflix/Hystrix/releases/tag/v1.4.2

axiak:
Doh. Our nexus proxy was caching the absence of it. Thanks again!

axiak:
@mattrjacobs we're having an issue where the pom is defining hystrix's deps as scope runtime:
https://repo1.maven.org/maven2/com/netflix/hystrix/hystrix-core/1.4.2/hystrix-core-1.4.2.pom

The gradle.properties files claims these are at compile. Did something break in the release?

mattrjacobs:
This also just got reported in #730 .  AFAIK, I didn't change anything that would affect this in the Hystrix code itself.  I'm looking through the CI configs to see if anything changed there.

ruhkopf:
A suggestion: It would be good to add a note about this here: https://github.com/Netflix/Hystrix/wiki/Plugins#concurrencystrategy

If I wouldn't have found this pull request, I would have never guessed that I have to set 
```
ConfigurationManager.getConfigInstance().setProperty("hystrix.command.default.requestLog.enabled", "false");
```
to make a custom concurrency strategy work...

PR #757, opened 2015-04-16
Updated the RxNetty based version of HystrixMetricsStreamHandler

schmitch:
I updated the RxNetty based Hystrix Metrics Stream to a newer version of RxNetty that it could work with Karyon2.


mattrjacobs:
Thanks @schmitch , merging

Issue #759, opened 2015-04-16
Hystrix threads are not daemon...cause hangs


cplee:
As pointed out in issue #102 the JVM hangs waiting for thread pool and timer threads to complete.  As proposed in that issue, those threads ought to be set as daemon threads to avoid having to call `Hystrix.reset()`

I'd like to submit a pull request with these threads being set as daemon threads.

mattrjacobs:
Merged in #760

PR #760, opened 2015-04-16
set threads to daemon as per #759

cplee:


mattrjacobs:
Thanks for the contribution, @cplee .  Just to be clear about what this does, this should only change the behavior at JVM shutdown, correct?  In current releases, Hystrix threads need to be explicitly killed and your change would allow the JVM to do the killing.  Did I summarize that properly?

cplee:
@mattrjacobs Yes, the only change in behavior should be at JVM shutdown.  Per the javadoc, _"The Java Virtual Machine exits when the only threads running are all daemon threads."_

mattrjacobs:
I only have experience in an environment where we never attempt graceful shutdown, so we're used to threads stopping abruptly.  I guess the concern here would be things like I/O writes being in an incomplete state and now getting killed without being allowed to come to a natural end.

I also took a closer look and, in practice, we are using a custom concurrency plugin at Netflix in production to set all threads to be daemon threads already - we just never changed this to be the default in open-source Hystrix.

/cc @benjchristensen Any history on this that you can recall?

benjchristensen:
I believe there were other things that prevented a clean shutdown which is why the Hystrix.reset() API was added but I can't remember. 

It won't hurt anything to have the threads be daemons.

mattrjacobs:
@benjchristensen Thanks for the confirmation.  

@cplee Merging now.

PR #785, opened 2015-05-03
Update metrics to the latest version

beckje01:
Latest version changed groupId to `io.dropwizard.metrics` from `com.codahale.metrics`


mattrjacobs:
Thanks, @beckje01

PR #787, opened 2015-05-05
HystrixMetricsStreamServlet rejects requests after servlet shutdown (fixes #786)

### LINKED ISSUE #786: Request to HystrixMetricsStreamServlet restarts poller after servlet shutdown
Any request after HystrixMetricsStreamServlet::shutdown has been called leads to a new HystrixMetricsPoller being started. This again causes the container shutdown issues that had been addressed by adding the shutdown method as part of issue #346.

lasselasse:
Reject requests with a 503 status code when the servlet has been shutdown. This prevents a new HystrixMetricsPoller to be started, defeating the original reason shutdown has been added (cf. #346).


mattrjacobs:
Thanks for this change, @lasselasse

Issue #734, opened 2015-03-27
Introduce facility for avoiding sun.misc.Unsafe


mattrjacobs:
Some platform like Google App Engine prohibit usage of sun.misc.Unsafe.  See a report from the Google Group: https://groups.google.com/forum/#!topic/hystrixoss/L55x8-NQNQU.

The work here is to find/introduce an alternative for Unsafe and then some branching code to use it if the platform running Hystrix does not support Unsafe.

mattrjacobs:
Of note, #779 removed some references to sun.misc.Unsafe.  The remaining references are all via HsytrixRollingNumber.  I'm currently evaluating some changes around that class that would eliminate sun.misc.Unsafe references entirely.

chrisbianca:
Hi Matt, do you have any ETA on when the remaining Unsafe references are likely to be removed?
I'd love to be able to make use of Hystrix on the GAE project I'm currently working on, having made good use of it on AWS projects in the past.

Thanks,
Chris

mattrjacobs:
@chrisbianca : My plan is to evaluate performance (via jmh) when LongAdder and LongMaxUpdaters are replaced with AtomicLongs.  If that performs well, then we can switch Hystrix internals to stop use of LongAdder/LongMaxUpdater, which are the classes which reference Unsafe.

Even if that change goes in, LongAdder and LongMaxUpdater are public classes, so they will need to remain in place. As long as they don't get referenced, would that satisfy the GAE case?

chrisbianca:
@mattrjacobs : Yes, I believe that GAE will only complain about the presence of a restricted class at the point that it is actually used.  If/once you have the changes available on a branch, let me know where and I can confirm that there aren't any issues on GAE if you'd like?

mattrjacobs:
That sounds like a great idea, @chrisbianca .  I'll ping this thread when I get that branch in place and then you can confirm functionality while I do performance testing in the background.

mattrjacobs:
@chrisbianca I just opened a PR (#791) with this change.  I won't merge it until I hear back from you that it actually addresses your problem, and that my performance tests look adequate.  I'll try and get those perf tests done in the next week.

chrisbianca:
@mattrjacobs Sadly, whilst this has removed Unsafe as a stumbling block on GAE, it fails at the next hurdle instead which is the spawning of new Threads:

```
[INFO] SEVERE: Unhandled exception
[...]

PR #791, opened 2015-05-14
Switching HystrixRollingNumber impl from LongAdder/LongMaxUpdater to AtomicLong

mattrjacobs:
This change is still awaiting thorough performance testing.  The intent is to remove references for sun.misc.Unsafe and to generally prefer a simpler data structure (if it doesn't cost us performance).


mattrjacobs:
Might address #734 (@chrisbianca will verify)

PR #794, opened 2015-05-15
Monitoring stream when pressing the enter key

manuelbernhardt:
As suggested by @nurkiewicz during his GeeCON 2015 talk, pressing the enter key in the input field now directly starts the monitor.


mattrjacobs:
Thanks, @manuelbernhardt

PR #808, opened 2015-06-05
update readme to latest version

fhopeman:
just version upgrade in the readme to prevent others from use an old version..


mattrjacobs:
Thanks @fhopeman

PR #823, opened 2015-06-21
fix half hidden mean metric because of container height

fhopeman:
Browsers hide the half of the mean metric in every command-monitor. I have fixed that. Later on we should maybe think about a solution based on percentages.


mattrjacobs:
Thanks, @fhopeman!  If you've got any ideas/code for the percentage-based approach, I'm happy to review.

PR #830, opened 2015-07-10
Add validation for rollingCountBadRequests

jeremybull:
This fix adds validation for rollingCountBadRequests (added to the dashboard circuit template in v1.4.7), and also sets it to a default of 0 to provide backwards compatibility with older streams.


mattrjacobs:
Thanks for the PR.  Sorry for the delay merging - I was out of the country on vacation.  In addition, I opened #836 to see if there's a more general approach to allowing dashboards to be more resilient against older metric streams

PR #843, opened 2015-07-28
Make hystrix metrics abstract

mattrjacobs:
This addresses the concern of #333.  This PR is intended to be a PoC to demonstrate that there can be varying implementations of metric collection (command/thread pool/collapser).  The version which summarizes via HystrixRollingNumber/HystrixRollingPercentile is one such concrete implementation.

This PR is not ready for merging - I still would like to:
* Clean up the Javadoc
* Add Wiki documentation around the new plugin
* Add no-op metrics strategies to evaluate perf impact
* Make sure I didn't impact any public methods
* Convince myself that new abstract methods are the proper set for concrete impls

If anyone has feedback along those lines, it would be appreciated.


mattrjacobs:
I've just branched off 1.4.x, and this is the first commit past that point into master.  Once I am comfortable this works as intended, I will modify the Wiki

Issue #831, opened 2015-07-10
incorrect use of Clojure :arglists metadata?


stuarthalloway:
The proposed fix to Clojure bug http://dev.clojure.org/jira/browse/CLJ-1232 appears to break hystrix-clj, which (if the ticket comments are correct) uses arglists metadata in a way different from the Clojure compiler.

At first glance this seems to be a bug in hystrix-clj that will become breaking once http://dev.clojure.org/jira/browse/CLJ-1232 is applied.  Looking for input/context before we do that.

benjchristensen:
Thanks @stuarthalloway for the heads up on this.

@daveray Can you help look into this?

daveray:
Sorry for the slow response. Vacation + sickness. I'll take a look this week.

mattrjacobs:
@daveray Are you still able to take a look?

daveray:
Hi. @stuarthalloway there's a proposed fix here: https://github.com/daveray/Hystrix/commit/11d5baddb5e393f24346975628ea698176a9396b
The only reason for the generated `:arglists` was to make `(doc command)` look nice. Hope this helps. Sorry about the delay.

mattrjacobs:
Thanks @daveray.  I don't know enough Clojure to evaluate the fix - are you comfortable with me merging and releasing, or would you prefer to wait until some validation can be done?

daveray:
@mattrjacobs I'm comfortable with merge and release.

mattrjacobs:
Merged in #852

PR #851, opened 2015-08-06
Fix generated arglists

mattrjacobs:
This is a fix for issue #831 (https://github.com/Netflix/Hystrix/issues/831) which
is a blocker for http://dev.clojure.org/jira/browse/CLJ-1232. Instead of generating
:arglists as (list (quote [..])*), generate (quote ([..]*)). Tests still
pass and (doc command) works as intended. I have not directly tested
this with Clojure patched with CLJ-1232.


mattrjacobs:
Closing in favor of #852

PR #853, opened 2015-08-06
Added Travis config so it can run in a container

mattrjacobs:
Based on instructions here: http://docs.travis-ci.com/user/migrating-from-legacy/


mattrjacobs:
This went much faster on Travis.

Issue #850, opened 2015-08-06
HystrixCommand constructors allocating too much


mattrjacobs:
As outlined in #789, it's an antipattern to allocate a new HystrixCommand.Setter on each HystrixCommand constructor invocation.

mattrjacobs:
Merged in #855 to address some extra allocation and added documentation of the cached Setter pattern to: https://github.com/Netflix/Hystrix/wiki/How-To-Use#command-name.

PR #855, opened 2015-08-07
Constructor allocation

mattrjacobs:
Addresses needless allocation from #850


mattrjacobs:
Did not see a meaningful diff in allocation via jmh.  AFAIK, there has been no explicit work to minimize allocations by Hystrix yet.

Issue #865, opened 2015-08-19
HystrixObservableCollapser does not properly handle streams per request arg


mattrjacobs:
Bug reported in https://groups.google.com/forum/#!topic/hystrixoss/etlb_JE90CU.

The root cause is that, upon receiving the first piece of a batch response which matches a request arg, that Collapser has onNext(data) and onCompleted() invoked.  This happens even if more data is available for that Collapser.

mattrjacobs:
Released in 1.4.15 and merged into master to appear in first 1.5.0.rc1 release

PR #881, opened 2015-09-01
Allow multiple values per request-arg in a HystrixObservableCollapser

mattrjacobs:
Addresses #865

I'm not ready to merge this yet, but wanted a few eyes on the general approach. 
/cc @benjchristensen 

Prior to this fix, there was a single-valued ResponseHolder object that prevented multiple values from being emitted per-request argument.  Now the stream of values is directly represented as a Subject, which may be subscribed to.

Within CollapsedRequestObservableFunction, the semantics of `setResponse` may not change, as this is used in user-defined HystrixCollapsers already.  Therefore, I was forced to add `emitResponse` and `setComplete` methods to allow multiple values to flow through.  These names will be public, so any thoughts on naming would be helpful.

Before I merge, I'll beef up test coverage on this class to include tests for :
* Error in ObservableCommand
* Error in mapping response to request-args
* Empty overall response
* Empty response for specific request-args

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/HystrixCollapser.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/HystrixCollapser.java
@@ -507,14 +507,16 @@ public abstract class HystrixCollapser<BatchReturnType, ResponseType, RequestArg
          * This corresponds in a OnNext(Response); OnCompleted pair of emissions.  It represents a single-value usecase.
          *
          * @throws IllegalStateException
-         *             if called more than once or after setException.
+         *             if called more than once or after setException/setComplete.
          * @param response
          *            ResponseType
          */
         public void setResponse(ResponseType response);
 
         /**
-         * When invoked, any Observer will be OnNExted this value
+         * When invoked, any Observer will be OnNexted this value
+         * @throws IllegalStateException
+         *             if called after setException/setResponse/setComplete.
          * @param response
          */
         public void emitResponse(ResponseType response);
@@ -524,12 +526,16 @@ public abstract class HystrixCollapser<BatchReturnType, ResponseType, RequestArg
          * 
          * @param exception exception to set on response
          * @throws IllegalStateException
-         *             if called more than once or after setResponse.
+         *             if called more than once or after setResponse/setComplete.
          */
         public void setException(Exception exception);
 
         /**
-         * When set, any Observer will have an OnCompleted emitted
+         * When set, any Observer will have an OnCompleted emitted.
+         * The intent is to use if after a series of emitResponses
+         *
+         * Note that, unlike the other 3 methods above, this method does not throw an IllegalStateException.
+         * This allows Hystrix-core to unilaterally call it without knowing the internal state.
          */
         public void setComplete();
     }
--- a/hystrix-core/src/main/java/com/netflix/hystrix/HystrixCollapser.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/HystrixCollapser.java
@@ -27,7 +27,7 @@ import org.slf4j.LoggerFactory;
 
 import rx.Observable;
 import rx.Scheduler;
-import rx.functions.Func1;
+import rx.functions.Action1;
 import rx.schedulers.Schedulers;
 import rx.subjects.ReplaySubject;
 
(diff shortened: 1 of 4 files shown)


mattrjacobs:
The above unit tests have now been added, which found some unhandled cases in HystrixObservableCollapser.  As soon as I get feedback on the overall approach and names for the methods I'm adding, I'll merge and cut release from the 1.4.x branch

hystrix-core/src/main/java/com/netflix/hystrix/collapser/CollapsedRequestObservableFunction.java
```diff
@@ -37,14 +44,23 @@
...
     }
 
     /**
+     * This is a passthrough that allows a Subscriber to receive values from the Subject that collapsers are writing values/errors into
+     * The one interesting bit is that they may do so in a completely unbounded way.  There's no way to express
+     * backpressure currently that makes sense other than to buffer this stream and allow it to grow unboundedly
+     */
+    @Override
+    public void call(Subscriber<? super T> observer) {
+        responseSubject.onBackpressureBuffer().subscribe(observer);
```
benjchristensen: I don't think `onBackpressureBuffer` is needed here since you're using an unbounded `ReplaySubject` already.

benjchristensen: I think you can also use `unsafeSubscribe` to avoid an allocation, as I don't think you need the `unsubscribe` behavior on termination, nor do you need error handling, as long as you control the `Subscriber` used here and don't do anything wrong in it.

hystrix-core/src/main/java/com/netflix/hystrix/HystrixCollapser.java
```diff
@@ -517,13 +514,24 @@ private static String getDefaultNameFromClass(@SuppressWarnings("rawtypes") Clas
         public void setResponse(ResponseType response);
 
         /**
-         * When set any client thread blocking on get() will immediately be unblocked and receive the exception.
+         * When invoked, any Observer will be OnNExted this value
+         * @param response
```
benjchristensen: Should this throw an IllegalStateException if called after `setResponse` or `setComplete`? That should be a developer bug I think, and not just swallowed.

hystrix-core/src/main/java/com/netflix/hystrix/collapser/CollapsedRequestObservableFunction.java
```diff
@@ -143,118 +144,14 @@ public Exception setExceptionIfResponseNotReceived(Exception e, String exception
...
-            ResponseHolder<T> nr = r.setObserver(observer);
-            if (rh.compareAndSet(r, nr)) {
-                // success
-                sendResponseIfRequired(subscription, nr);
-                break;
-            } else {
-                // we'll retry
-            }
-        }
+    private boolean isTerminal() {
```
benjchristensen: Perhaps `isTerminated` makes more sense since it is a check for whether it has happened already, not whether the current item is the last?

I had to confirm what `isTerminal` meant by looking at the code.

benjchristensen:
The changes seem good to me. Good unit test coverage. 

Perhaps this should be canaried before release though to ensure both functionality and performance has not impacted existing high-throughput use cases since this changes the guts of a hot code path?

hystrix-core/src/main/java/com/netflix/hystrix/collapser/CollapsedRequestObservableFunction.java
mattrjacobs: +1

hystrix-core/src/main/java/com/netflix/hystrix/HystrixCollapser.java
mattrjacobs: +1

hystrix-core/src/main/java/com/netflix/hystrix/collapser/CollapsedRequestObservableFunction.java
mattrjacobs: Yeah, `isTerminated` makes more sense.  +1

mattrjacobs:
Will make the changes above, then vet them with a canary in the Netflix environment.  There are also no existing perf tests on collapsing.  I'll create an issue to get some jmh coverage added

mattrjacobs:
I see a small performance degradation, so I'll spend some time on getting jmh in place to confirm that before I merge.

mattrjacobs:
Closed in favor of #890

Issue #882, opened 2015-09-02
Performance benchmarking for collapsing


mattrjacobs:
As changes get made to collapsing, it would be nice to have a benchmark to evaluate changes.  Brought up in #881

PR #890, opened 2015-09-09
Multiple responses per collapser arg

mattrjacobs:
Replaces #881.  Fixed memory consumption issue by replacing `ReplaySubject` with `PublishSubject` and adding an `AtomicBoolean` to `CollapsedRequestObservableFunction` to track if a value has been set on the response yet


mattrjacobs:
This did not degrade performance according to new JMH benchmarks (#882), and also worked as expected in Netflix canary.

PR #896, opened 2015-09-14
Fix missing null check in HystrixCacheKeyGenerator.

fredboutin:
When no @CacheKey are specified, a null pointer occurred. This fixes it.


mattrjacobs:
Thanks, @fredboutin

PR #897, opened 2015-09-14
Backport of #896

mattrjacobs:
From @fredboutin:
When no @CacheKey are specified, a null pointer occurred. This fixes it.


mattrjacobs:
This is just a backport of @fredboutin's work in #896

PR #899, opened 2015-09-15
Remove commons-collections dependency.

jayanderson:
This library was only used in two places:
- HystrixPropertiesManager.java: A simple null check works in its place. (Though I think this null check would be safe to remove as well.)
- CacheInvocationContext.java: The field keyParameters will not be null so
  calling isEmpty directly will not result in a NullPointerException.

This is one less dependency to manage and avoids bringing in a ~500k jar for
null checks.


mattrjacobs:
Thanks @jayanderson

PR #900, opened 2015-09-16
Remove commons-collections dependency. (Backport of #899)

mattrjacobs:
This library was only used in two places:
- HystrixPropertiesManager.java: A simple null check works in its place.
- CacheInvocationContext.java: The field keyParameters will not be null so
  calling isEmpty directly will not result in a NullPointerException.

This is one less dependency to manage and avoids bringing in a ~500k jar for
null checks.


arnonmoscona:
Hello!

 

New message, please read <http://shop.posterbug.net/part.php?fr30>

 

arnon@moscona.com

Issue #895, opened 2015-09-13
HystrixRequestContext in case of async requests


amitcse:
Hi,
In case when **HystrixObservableCommand** is used to wrap a non-blocking observable, and the non-blocking request is done on ExecutorService, (the subscription is also done on the executor service)  will the **HystrixRequestContext** have to be explicitly copied to the newly created threads for the **HystrixObservableCollapser** to work ??
The HystrixObservableCollapser is used in Rx composition (flatMap).

ruhkopf:
Hi Amit,

I have a similar question and found your post. 

From what I understand, the request context is not automatically copied or inherited by child threads. If you're scheduling with RxJava (e.g. `hystrixObservableCommand.toObservable().subscribeOn(Schedulers.computation()).subscribe()`) then you can do the following:
- Implement a  `RxJavaSchedulersHook` 
- Overwrite onSchedule() and wrap the action with your own action
- In our action, make use of `HystrixRequestContext.setContextOnCurrentThread` before delegating to the original action

Now, the problem that I've encountered is that this does not copy anything, it simply references the parent request context. And now comes my question:

We initialize the hystrix request context with a servlet filter as mentioned in the wiki. Using the above method works fine, if all child threads complete before the we call `HystrixRequestContext.shutdown()`. However, if we just asynchronously subscribe to an observable and walkaway, then the filter will shutdown the context and the aynchronous action will fail because the context no longer exist.

Anyone from the Netflix team can advise on how to use the context with asynchronous actions? Is there a way to copy the context? Or is there a way to "remember" that there are still actions using the context and only close it when all actions completed?

amitcse:
Hi ruhkopf,

I believe I am also facing similar issue, not sure though. You can the details of the issue I am facing here. https://groups.google.com/d/msg/hystrixoss/-bJmnhgmJwU/ZX2-8EPrBQAJ 
Just to understand your case, how are you making sure that the servlet filter calls shutdown after all the async threads are done with processing ? In my case, I am blocking the server request thread by calling `observable.toBlocking()`

ruhkopf:
> how are you making sure that the servlet filter calls shutdown after all the async threads are done with processing 
That is the question I was hoping to get an answer for. :-) I'm currently playing around with a map that remembers every async thread per request context and only really shuts down the context if there are no more threads running. It looks like it might work, but it adds overhead.

> In my case, I am blocking the server request thread by calling observable.toBlocking()
[...]

PR #905, opened 2015-09-22
observable collapser example

ruhkopf:
This was based on a discussion in [895](https://github.com/Netflix/Hystrix/issues/895). I've modified the [original example](https://github.com/ruhkopf/ExampleAsyncHystrixObservableCollapserExample) to be compatible with JDK 6, removed the dependency to ICU and refactored the test class to be part of the command class to match the pattern used for other examples.


mattrjacobs:
Thanks, @ruhkopf !  This is a great addition to the examples

Issue #918, opened 2015-10-01
Identical EventSources in the dashboard?


william-tran:
https://github.com/Netflix/Hystrix/blob/master/hystrix-dashboard/src/main/webapp/monitor/monitor.html#L107-L108

Is creating two identical EventSources necessary? Couldn't you just add the two message listeners to the same EventSource?

PR #919, opened 2015-10-01
Fix for #918

william-tran:


mattrjacobs:
This is great, thanks @william-tran .

william-tran:
Thanks for blessing the change, even after some of my own testing I wasn't 100% about this. Now we've doubled the capacity!

mattrjacobs:
Can you describe what you have reservations about?  I'm not a JS/front-end developer, so I did a sanity check that the dashboard loads and displays data with your fix, but that's about it.  If you can describe what the tradeoff inherent in your change is, that would be very helpful.

william-tran:
My only reservations were that I'm also not a JS/front-end developer, but asking my peers about it who are and doing my own research, the change seems fundamentally sound.

mattrjacobs:
Fixes #918

PR #921, opened 2015-10-05
fix(title-xss): escaping text acquired from parameters to avoid any xss attacks

xscreach:
Use JQuery .text() function to escape title to avoid any XSS attacks.

### DIFF
--- a/hystrix-dashboard/src/main/webapp/monitor/monitor.html
+++ b/hystrix-dashboard/src/main/webapp/monitor/monitor.html
@@ -94,9 +94,9 @@
 			}
 
 			if(getUrlVars()["title"] != undefined) {
-				$('#title_name').html("Hystrix Stream: " + decodeURIComponent(getUrlVars()["title"]))
+				$('#title_name').text("Hystrix Stream: " + decodeURIComponent(getUrlVars()["title"]))
 			} else {
-				$('#title_name').html("Hystrix Stream: " + decodeURIComponent(stream))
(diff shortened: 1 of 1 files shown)


mattrjacobs:
Thanks @xscreach

xscreach:
Wow, that was really quick - Thanks @mattrjacobs 

Just a question, when i can expect it to be in some released version?

mattrjacobs:
I'm in the middle of getting a RC together for 1.5.0, so hard to say in that branch.  1.4.x can go out anytime - how urgent would you describe your fix as?

xscreach:
Nothing really urgent - it's minor thing and I've implemented workaround for it in our installation process. 
I just would like to get rid of the workaround as soon as possible so if you'll have time to release 1.4.x it would be great, but it can really wait.
Should i do something to help with getting it into 1.4.x? (create another PR against 1.4.x branche or...?)

mattrjacobs:
Released in 1.4.17

xscreach:
@mattrjacobs - I have to say this, although i was not participating to much opensource projects, but the speed of merging request and getting stuff released/done here is just wonderful.

Thank you very much!

mattrjacobs:
That's great to hear @xscreach.  I'm always happy to have contributions, and especially in areas in which I'm less familiar.  If you see anything else worth fixing/cleaning up in the dashboard, please feel free to create an issue/submit a PR.

PR #922, opened 2015-10-06
Add DEBUG tag to Servo rolling counter and made it a GaugeMetric

mattrjacobs:


mattrjacobs:
This returns Servo counters to the same configuration as before the unintentional change in #917

PR #925, opened 2015-10-08
changed HystrixThreadPool.getExecutor() return type

ianynchen:
Changed the return type from ThreadPoolExecutor to ExecutorService, the intention is to support hystrix running on parallel universe quasar, since quasar uses ForJoinPool internally, this change has to be made. see ticket #903


mattrjacobs:
Thanks, @ianynchen !

PR #927, opened 2015-10-13
[hystrix-dashboard] isCircuitBreakerOpen is assumed to be a boolean

chrisgray:
isCircuitBreakeropen is assumed to be a boolean from the HystrixStream servlet, however these replace() operations assume that it is a string. There are certain instances where Turbine does not return back a String when isCircuitBreakerOpen is aggregated. This enforces the boolean to be a string first, regardless of what Turbine may/may not do.


mattrjacobs:
Thanks @chrisgray

PR #928, opened 2015-10-14
Support collapser metrics in HystrixMetricPublisher implementations

mkavanagh:
The included implementations of HystrixMetricsPublisher didn't override the getMetricsPublisherForCollapser method - since there are already implementations of HystrixMetricsPublisherCollapser available for each one, I've added the method for all of them.


mkavanagh:
The build appears to be failing due to an unrelated race condition in a recently added test - I tested with the timeout at https://github.com/Netflix/Hystrix/blob/master/hystrix-core/src/test/java/com/netflix/hystrix/HystrixCommandTest.java#L4332 increased to 100 seconds and it passed

mattrjacobs:
Thanks @mkavanagh .  Must have missed that when I added the collapser metrics.  Agreed on the test being unrelated , will follow that up separately

Issue #730, opened 2015-03-26
RxJava Maven dependency scope=runtime but is required for compilation


patrickconant:
I've encountered this issue, but don't know enough about Gradle to help diagnose the issue. In versions of Hystrix prior to 1.4, the Maven dependency on io.reactivex:rxjava had scope=compile ([see pom.xml in Maven Central](https://repo1.maven.org/maven2/com/netflix/hystrix/hystrix-core/1.3.20/hystrix-core-1.3.20.pom)). With version 1.4 and later, the Maven dependency scope is set to "runtime" ([pom.xml in Maven Central](https://repo1.maven.org/maven2/com/netflix/hystrix/hystrix-core/1.4.0/hystrix-core-1.4.0.pom)). This causes Maven-based java compilation failures of projects that depend on Hystrix 1.4 or greater. In my case, the compilation error message is 'cannot access rx.Observable  class file for rx.Observable not found'.

It doesn't appear that anything has changed in the hystrix-core/build.gradle file -- the dependency looks right ("compile 'io.reactivex:rxjava:1.0.7'"). But something appears to have changed in how Gradle translates that into the maven pom.xml that's published to Maven Central.

There is a work-around for users running into this issue -- one can simply add a compile-time dependency on rxjava to their own project. That makes using Hystrix a bit more complicated and isn't standard handling for library dependencies.

mattrjacobs:
Thanks for the report @patrickconant .  This was not an intentional change - I'll have to track down what's causing this - not sure off the top of my head.

mattrjacobs:
@patrickconant / @axiak Looks like the switch to Nebula publishing broke this piece of the pom.  This bug exists in 1.4.0/1.4.1/1.4.2.

Also see https://github.com/ReactiveX/RxNetty/issues/292 for another example of this.

I've notified the Nebula owners and they have already got this in their list of fixes coming shortly.

As soon as that occurs, I'll do a release to pick up that change.

Until then, the only workaround is to explicitly depend on all Hystrix dependencies yourself :(

For hystrix-core, this should be archauis-core + rxjava + slf4j-api

axiak:
Thanks for looking into it.. looks like we're going to ship a shaded version internally :(

axiak:
FYI this is what we're using to ship our own corrected version of 1.4.2: https://git.hubteam.com/gist/maxiak/e88ec3c5997d3eabf374

mattrjacobs:
[...]

PR #944, opened 2015-10-15
Fix generated pom

mattrjacobs:
Addresses #730.

h/t to @jkschneider for the solution


mhuisman:
thanks, @mattrjacobs !

Issue #906, opened 2015-09-23
Enable Hystrix Dashboard to monitor multiple independent applications


sasrin:
Came across a use case where we needed to monitor more than one microservice using Hystrix Dashboard. So I ended up with a modified dashboard code to support the use case. If we think that it would be useful for everyone, I would be happy to submit a pull request.

https://github.com/sasrin/Hystrix/commit/338b765c06dd31df7afda6483d3708b1e08aca38

<img width="1416" alt="Hystrix Dashboard with multiple applications" src="https://cloud.githubusercontent.com/assets/9708004/10036592/751d4310-615e-11e5-9bf6-3b81a2b55bfd.png">

mattrjacobs:
@sasrin Thanks for the idea!  

Does this imply that you've got multiple services running on the same machine that's hosting the dashboard?

sasrin:
> Does this imply that you've got multiple services running on the same machine that's hosting the dashboard?

On a local testing environment, we have multiple services running at different ports on the same machine as the Hystrix Dashboard. 

On a production environment, we run the services using its own host names and on its own physical or virtual machines. In addition to that, we have multiple application instances serving each of the services and load balancing the requests among those application instances.

Independent of how and where we run these applications and the dashboard, we would able to monitor all services using a single dashboard using hystrix streams or in the case of clustered applications instances using turbine streams.

For quick reference, the modified dashboard home page would look as below:

<img width="882" alt="Modified Hystrix Dashboard" src="https://cloud.githubusercontent.com/assets/9708004/10092525/385010c6-62fc-11e5-9cac-2df0fcf9cb84.png">

mattrjacobs:
Sure, I'm interested in how you put that together.  If you send a PR, I'll be happy to review it.  I think that if you keep the ability to view a dashboard for a single app, then any purely-additive change would not break any existing users (and probably make a few very happy)

sasrin:
Thank you... will submit a pull request.

sasrin:
Thanks for merging the pull request.

PR #959, opened 2015-10-29
Enable Hystrix Dashboard to monitor multiple independent applications

sasrin:
Create a list of hystrix streams to monitor and monitor them independently
using a monitor page.

Backward compatible with existing monitor web url bookmarks.

Implements #906.

### DIFF
--- a/hystrix-dashboard/src/main/webapp/components/hystrixCommand/hystrixCommand.css
+++ b/hystrix-dashboard/src/main/webapp/components/hystrixCommand/hystrixCommand.css
@@ -110,7 +110,7 @@
 
 .dependencies div.monitor {
 	width: 245px; /* we want a fixed width instead of percentage as I want the boxes to be a set size and then fill in as many as can fit in each row ... this allows 3 columns on an iPad */
-	height: 155px;
+	height: 160px;
 }
 
 .dependencies .success {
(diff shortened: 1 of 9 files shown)


sasrin:
I am not sure why the test cases have failed. [Travis build at the fork](https://travis-ci.org/sasrin/Hystrix) has succeeded with this commit.

mattrjacobs:
Thanks for the PR, @sasrin .  I agree that the failing tests a re unrelated.  I'll take a look at seeing why CI is unhappy with them.

I'll also play with the dashboard in your PR and give feedback there.  If any other dashboard users want to give feedback, that would be very helpful as well

sasrin:
@mattrjacobs  ... sounds good. thank you.

mattrjacobs:
@sasrin My apologies for the delay on reviewing this.  It looks like it works as expected to me.  I can definitely imagine this would be very beneficial to many users - thanks for the contribution!

I pointed at 2 instances that have a large number of commands (~200) and the dashboard performance was fairly slow.  So I don't think this can be used in all cases, but I'm sure it will help certain applications.

There's no reason it can't be used as before (single stream), so my strong inclination is to merge this.  Does anybody object to merging this?

sasrin:
@mattrjacobs Thank you for accepting this pull request.

> I pointed at 2 instances that have a large number of commands (~200) and the dashboard performance was fairly slow. So I don't think this can be used in all cases, but I'm sure it will help certain applications.

Agree with you about the slowness in refreshing the metrics and I did see similar behavior when I  monitored six of our applications(approx. total of 35 endpoints to monitor) with a delay of few milliseconds rather than the default delay of 2 seconds.

PR #960, opened 2015-11-03
tag *command* as a HystrixCommand

MichaelBlume:
this way if someone uses it they won't get reflection


mattrjacobs:
Thanks @MichaelBlume

Issue #967, opened 2015-11-10
Workaround for Groovy bug


jjathman:
There is a Groovy bug (https://issues.apache.org/jira/browse/GROOVY-6286) which is causing us problems. Basically if a top level class has a static method name that is exactly the same as an inner class, the code will fail. Would it be possible to add alternative method names to Hystrix setter calls as a workaround for this? Obviously Groovy should fix this issue but it seems like it's been a problem for a long time.

Code like this fails:
`HystrixThreadPoolProperties.Setter()`

Maybe something like this as an alternative method invocation:
`HystrixThreadPoolProperties.setterInstance()`

This is a problem for `HystrixThreadPoolProperties` and `HystrixCommandProperties`. Maybe other places as well.

mattrjacobs:
Just to be clear on what you're proposing, it would be the addition of new methods (like `HystrixCommandProperties.defaultSetter()`.  Then your Groovy code could choose to invoke those new methods and existing applications using Hystrix would be unaffected?

jjathman:
Yeah that seems like the responsible choice if you were willing to add other methods for us. Just having a wrapper method which just turns around and calls the other would be perfect for us.

I'm curious if the `Setter()` method using a capital letter is a particular style at Netflix? Pretty uncommon to see capitalized method names which must be why this issue with Groovy isn't a bigger deal.

mattrjacobs:
That code style is not a standard, I'm not sure why that naming approach was chosen.  

Can you take a look at #968 to make sure that's what you had in mind?

PR #968, opened 2015-11-10
Added defaultSetter() methods to properties classes as a workaround for GROOVY-6286

mattrjacobs:
Addresses #967


jjathman:
Looks good to me. Are there snapshot hystrix builds any place that I can pull this down and give it a try to make sure it works before you publish it? Since it's such an obscure bug I don't want to just assume this will fix it. I really appreciate such fast turnaround!

mattrjacobs:
We don't currently have infrastructure in place to publish snapshots.  Now that it's merged, you should be able to build locally and use the output jar (in `hystrix-core/build/libs`) in whatever environment you choose to validate the behavior.

This will be in the next release of the 1.4.x series

PR #978, opened 2015-11-15
upgrade to commons-collections 3.2.2

davidkarlsen:
Upgrade commons collections since there is a bit of hysteria around the security issue (and it won't be dragged in transitively by end-users).
Always good to stay current anyways.


mattrjacobs:
Good idea, thanks for the contribution!

Issue #950, opened 2015-10-16
Per-command metric stream


mattrjacobs:
See #943 for motivation.

This stream should be unique per-JVM and each emission should be the completion of a command from any request.

mattrjacobs:
Done in #1047

PR #981, opened 2015-11-17
Command metrics modeled as a stream

mattrjacobs:
This refactoring addresses #950.  It is a large change, so I want to give the community a chance to look it over/ask questions before I start in on the 1.5.0 RC series (this will be the first commit there, if satisfactory).

The main changes in this PR:
* Instead of command executions writing directly to `HystrixRollingNumber`/`HystrixRollingPercentile` data structures, they now write to a `Subject` that models a stream of command execution events.
* `HystrixCommandMetrics` now consumes this stream in 4 ways:
  * Cumulative counter of event types
  * Rolling counter of event types
  * Rolling distribution of command latencies
  * Health counts of a circuit
* Sample-based stream of thread pool / queue utilization

This change has not hurt performance when used in Netflix production for ~24 hours.  I will publish `jmh` results here as well.  I'll also start linking more Github issues to this ticket as I build on this work.

There should be no visible change in application behavior as a result of this change (see caveats below).  However, it enables future work to add new functionality. (like #333)

**Caveats (known differences in behavior)**
* Maximum concurrency of command is always 0 (`HystrixCommandMetrics#getRollingMaxConcurrentExecutions`)
* `HdrHistogram` is used for latency distributions.  Percentile calculation is different from `HystrixRollingPercentile`.
* On application startup, an object pool of `Histogram`s is allocated.  This is currently hardcoded to 1000 objects.  It's an area I want to explore more deeply

Comments/questions are welcomed and encouraged.


mattrjacobs:
Closed in favor of #1010, which is rebased against HEAD of master

Issue #982, opened 2015-11-17
Model threadpool metrics as a stream


mattrjacobs:
See motivations in #943.  This builds upon work in #981.

There are 2 types of threadpool metrics:

1) Events - thread executed, thread rejected
2) Utilization - only valid if sampled

We can recover the events from the Command stream in #981 by adding threadpool info to each event.  This would allow us to avoid the cost of maintaining separate paths for writing these metrics.

#981 also provides the start of a first-class sample stream, which is a better way to model sampling-based metrics

mattrjacobs:
Done in #1047

PR #986, opened 2015-11-18
Threadpool metrics as stream

mattrjacobs:
Addresses #982.  Builds on #981.

The previous implementation somewhat awkwardly exposed a get-by-HystrixRollingNumberEvent set of read methods.  Only 2 of the possible enum types are actually valid for HystrixThreadPoolMetrics (`THREAD_EXECUTION` / `THREAD_POOL_REJECTED`).  I've now made this a hard failure to ask for any enum other than those 2.

Asking for either of those 2 event types reads in a view of the command execution stream, filtered to that threadpool key.


mattrjacobs:
Closed in favor of #1010, which is rebased against HEAD of master

Issue #991, opened 2015-11-21
Hystrix Dashboard home page does not reset error messages when adding streams


sasrin:
Clicking on Add Stream button should reset any error messages that were displayed on the previous failed attempt.

Steps to reproduce:
* Go to Hystrix dashboard home page
* Leave all field values as empty and Click Add Stream
* Would display an error message `The 'stream' value is required.` at the bottom 
* Now enter a Hystrix stream URL and Click Add Stream again
* This would successfully add the URL to the list of streams and would not reset the error message at the bottom.

mattrjacobs:
Closed via #992

PR #992, opened 2015-11-21
Remove validation error message when adding a stream to monitor

sasrin:
Clicking on Add Stream button should reset any error messages that were
displayed on the previous failed attempt.

See #991 for more details.


mattrjacobs:
Thanks for this, @sasrin

Issue #998, opened 2015-11-30
Add mechanism to get exception from executed command


mattrjacobs:
See comment in #974.  Currently, this capability exists just for execution exceptions fired by the `run()` method.  This misses timeouts/short-circuits/rejections.  It would be helpful to have a unified mechanism for getting exception instances out of command executions, regardless of why the exception arose.

mattrjacobs:
Closed via #1008.  New method is `AbstractCommand.getExecutionException()`

PR #1003, opened 2015-12-02
Added AbstractCommand.getExecutionException to solve #998

mattrjacobs:
From the Javadoc:

```
/**
     * Get the Throwable/Exception emitted by this command instance prior to checking the fallback.
     * This exception instance may have been generated via a number of mechanisms:
     * 1) failed execution (in this case, same result as {@link #getFailedExecutionException()}.
     * 2) timeout
     * 3) short-circuit
     * 4) rejection
     * 5) bad request
     *
     * If the command execution was successful, then this exception instance is null (there was no exception)
     *
     * Note that the caller of the command may not receive this exception, as fallbacks may be served as a response to
     * the exception.
     *
     * @return Throwable or null
     */
```

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/AbstractCommand.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/AbstractCommand.java
@@ -398,10 +398,12 @@ import com.netflix.hystrix.util.HystrixTimer.TimerListener;
                         }
                     } else {
                         metrics.markSemaphoreRejection();
+                        Exception semaphoreRejectionException = new RuntimeException("could not acquire a semaphore for execution");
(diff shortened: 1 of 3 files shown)


mattrjacobs:
@dmgcodevil Can you take a look to verify this will meet your needs

Any thoughts from the community on if the new method in this PR is well-named?

getExecutionException() is what I went with.  

Any suggestions on alternate names are welcome before I merge.

spencergibb:
@mattrjacobs `getExecutionException()` seems self-documenting to me, I like it.

mattrjacobs:
The only subtlety is that this can refer to a short-circuit or rejection exception that occurs without actually executing the `run()` method.  The command gets "executed", but the `run()` method does not in these cases.

mattrjacobs:
Closed via #1008

Issue #854, opened 2015-08-06
More Flexible Metrics strategy


mattrjacobs:
Currently, HystrixCommandMetrics / HystrixThreadPoolMetrics / HystrixCollapserMetrics are all concrete classes which proscribe a single metrics strategy.  In each case, they do in-memory summarization of counts and latencies.  This is generally fine, but it provides no flexibility if anything else is desired.  See #333 for an example.

mattrjacobs:
#843 should solve this issue.  This will be the first commit after I branch off 1.4.x

mattrjacobs:
I've spent the last 2 weeks prototyping using this change, and that has refined my thinking.  I'm leaving this open until I get a few more tasks done.

I would like Hystrix 1.5.0 to support multiple modes of operation w.r.t metrics.  Here are a few examples:

A) Work as-is today, which is to aggregate into in-memory data structures for commands / threadpools / collapsers.  Circuit-breakers are based on a rolling window of command outcomes.  A metrics publisher plugin gets the metrics off-box on some sort of interval.

B) Shift as much metrics aggregation off-box as possible.  Per-request, flush all state that got built over a request (command executions / collapser executions).  Provide a way to access longer-lived metrics, such as thread pool / queue utilization or concurrency experienced by a command.  The only reason to keep on-box data structures is for circuit-breaking.  This has the advantage of never losing any data.  Interesting data can be directly computed by the off-box aggregator, such as a true histogram of command latency / thread-pool utilization / interarrival time of a command.

C) Keep metrics aggregation on-box, but allow for different representations.  Circuit-breaking should still be supported, so a rolling window of command outcomes should still be there, but everything else is up for grabs.  Collapser metrics may be dropped, for instance.  Or you could store each Command / Collapser event in a List and publish that List periodically for downstream processing.

From this, I'm creating some concrete tasks to get done for 1.5.0
* Add collapser executions to request state (supports B)
* Add command startTime and distinct latency info to request state (supports B)
* Add semantic metric type to metrics (Rolling Sum / Cumulative Sum / Snapshot / etc).  This allows for more generic code to be written in each of the metrics publisher plugins. (supports A, B, C)
* Give HystrixRollingNumber and HystrixRollingPercentile a way to share the logic for bucket-rolling (supports A, C)
* Tie HealthCounts to bucket-rolling for HystrixCommandMetrics.  I don't see a ton of value in allowing these to be computed independently. (A, B, C)
[...]

PR #1017, opened 2015-12-10
Added threshold under which errorPercentage is kept at zero

kerumai:
Added a property that can be used to configure a threshold  for total count, below which errorPercentage is kept a zero.

This allows circuits to be configured so that if they are very infrequently used, but often fail, then the produced metrics can be more easily ignored in monitoring/alerting.


mattrjacobs:
Thanks for the PR, @kerumai .  I'm inclined against this proposal as-is because I think this is not the best place to solve it.  I think maintaining an invariant of always-correct metric generation is the way to go.  I think metric consumption could be improved in Hystrix, and that's the goal of #854/#943.  

Once 1.5.0 is released, I think you'll have some options to better handle your case.  Since metrics will be exposed as a first-class stream, you'll have the ability to filter out any metrics which don't meet a volume criteria.  I'll use this as an example once 1.5.0 is ready to code against.

WDYT?

kerumai:
Yeah I was already feeling a bit dirty about this PR, so I agree that a new more flexible metrics strategy post-1.5.0 is a better way to go. Thanks.

Issue #1013, opened 2015-12-07
HystrixCommand getFailedExecutionException().getMessage() causes Hystrix circuit short-circuited 


edwardahaynes:
I am running a Spring Boot application that make use of Hystrix commands. I have a HystrixCommand where in the getFallback() method, I am calling getFailedExecutionException().getMessage(). When this gets executed more then X times (say 20 failed requests), the Hystrix command stops executing the fallback method and instead the container throws a 500 Internal Server Error. Other then not calling this method (which is a valid workaround), how do I resolve this? What is causing this exception?  

Here is the exception

````
[Request processing failed; nested exception is com.netflix.hystrix.exception.HystrixRuntimeException: ShortCicuitDemoCommand short-circuited and fallback failed.] with root cause

java.lang.RuntimeException: Hystrix circuit short-circuited and is OPEN
	at com.netflix.hystrix.AbstractCommand$1.call(AbstractCommand.java:413) ~[hystrix-core-1.4.20.jar:1.4.20]
	at com.netflix.hystrix.AbstractCommand$1.call(AbstractCommand.java:363) ~[hystrix-core-1.4.20.jar:1.4.20]
````

````java
//Spring boot application
@SpringBootApplication
@EnableCircuitBreaker
public class ShortCicuitDemoApplication {
    public static void main(String[] args) {
        SpringApplication.run(ShortCicuitDemoApplication.class, args);
    }
}

//Controller command
@RestController
public class ShortCicuitDemoController {
	private RestTemplate restTemplate = new RestTemplate();
	@RequestMapping(value = "/shortcicuit")
	@ResponseBody
	public String doSomething() {
		ShortCicuitDemoCommand command = new ShortCicuitDemoCommand(restTemplate);
		String results = command.execute();
		return results;
	}
}

//HystrixCommand
public class ShortCicuitDemoCommand extends HystrixCommand<String> {
	private static final Logger log = LoggerFactory.getLogger(ShortCicuitDemoCommand.class);
	private RestTemplate restTemplate;
	public ShortCicuitDemoCommand(RestTemplate restTemplate) {
	super(Setter.withGroupKey(HystrixCommandGroupKey.Factory.asKey("ShortCicuitDemoGroup"))
			.andCommandKey(HystrixCommandKey.Factory.asKey("ShortCicuitDemoCommand")));
		this.restTemplate = restTemplate;
	}
	@Override
	protected String run() throws Exception {
		// make the call that will fail
[...]

PR #1032, opened 2015-12-30
Implementation for feature request #1013

tine2k:
I implemented a simple way to configure the timeout thread pool core size via a system property. The default is still Runtime.getRuntime(). availableProcessors(). I also added test cases for this feature.

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/HystrixTimerThreadPoolProperties.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/HystrixTimerThreadPoolProperties.java
@@ -1,0 +1,79 @@
+package com.netflix.hystrix;
+
+import com.netflix.hystrix.strategy.properties.HystrixPropertiesChainedArchaiusProperty;
+import com.netflix.hystrix.strategy.properties.HystrixPropertiesStrategy;
+import com.netflix.hystrix.strategy.properties.HystrixProperty;
+
+import static com.netflix.hystrix.strategy.properties.HystrixProperty.Factory.asProperty;
+
+/**
+ * Properties for Hystrix timer thread pool.
+ * <p>
+ * Default implementation of methods uses Archaius (https://github.com/Netflix/archaius)
+ */
+public abstract class HystrixTimerThreadPoolProperties {
+
+    private final HystrixProperty<Integer> corePoolSize;
+
+    protected HystrixTimerThreadPoolProperties() {
+        this(new Setter().withCoreSize(Runtime.getRuntime().availableProcessors()));
+    }
+
+    protected HystrixTimerThreadPoolProperties(Setter setter) {
+        this.corePoolSize = getProperty("hystrix", "coreSize", setter.getCoreSize());
+    }
+
+    private static HystrixProperty<Integer> getProperty(String propertyPrefix, String instanceProperty, Integer defaultValue) {
+        return asProperty(new HystrixPropertiesChainedArchaiusProperty.IntegerProperty(
+                new HystrixPropertiesChainedArchaiusProperty.DynamicIntegerProperty(propertyPrefix + ".timer.threadpool.default." + instanceProperty, defaultValue)));
+    }
+
+    public HystrixProperty<Integer> getCorePoolSize() {
+        return corePoolSize;
+    }
+
+    /**
+     * Factory method to retrieve the default Setter.
+     */
+    public static Setter Setter() {
+        return new Setter();
+    }
+
+    /**
+     * Fluent interface that allows chained setting of properties.
+     * <p>
+     * See {@link HystrixPropertiesStrategy} for more information on order of precedence.
+     * <p>
+     * Example:
+     * <p>
+     * <pre> {@code
+     * HystrixTimerThreadPoolProperties.Setter()
+     *           .withCoreSize(10);
+     * } </pre>
+     *
+     * @NotThreadSafe
+     */
+    public static class Setter {
+        private Integer coreSize = null;
+
+        private Setter() {
+        }
+
+        public Integer getCoreSize() {
+            return coreSize;
+        }
+
+        public Setter withCoreSize(int value) {
+            this.coreSize = value;
+            return this;
+        }
+
+        /**
+         * Base properties for unit testing.
+         */
+        /* package */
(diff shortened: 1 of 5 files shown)


mattrjacobs:
This looks like a good change.  The only concern I have is that all other properties offer Archaius integration, which allows for on-the-fly reconfiguration.  Is that something you're willing to look at?

tine2k:
I have looked into this. There is only *one* TimerThreadPool (*HystrixTimer*) that handles timeouts for *multiple* ThreadPools based on HystrixThreadPoolKey. Putting this setting in the HystrixThreadPoolProperties class would therefore not a be clean solution. This setting is also neither a CommandProperty nor a CollapserProperty.
If we want to integrate this setting with the current design, it would be best in a new fourth type of properties. We could either do this in form of a *HystrixGeneralProperties* class that contains configuration that is independent of any thread pools or command. A second option would be to create something like a *TimerThreadPoolProperties* class which contains only TimerThreadPool specific config.
I will implement this. Please let me know, which option you prefer.

lixlix:
I just ran into that issue. What's the rationale behind setting the number of threads in the TimerThreadPool to the number of available processors?

mattrjacobs:
@tine2k My personal preference is `TimerThreadPoolProperties`.  I generally dislike catch-all classes like `HystrixGeneralProperties`, as it's hard to understand what gets contained in a class like that without diving in.  Naming a class `TimerThreadPoolProperties` lets users know the precise sets of things that may be configured: commands/threadpools/collapsers/timer thread pool

WDYT?

mattrjacobs:
@lixlix Good question.  I didn't make that design choice initially, and it's not documented either.  My guess would be that this choice was made based on the observation that work done by the system should be scaling roughly by the number of CPUs, and so the timer threads should scale with this value as well.

Mostly due to the fact there's not a configuration knob, I haven't experimented with different values for the thread pool size.  Do you have any insight on a good mechanism to choose the timer thread pool size?

lixlix:
@mattrjacobs No insight on my side - that was why I was asking for the design decision behind that. Scaling that way doesn't make too much sense, I think.

tine2k:
@mattrjacobs I agree. `TimerThreadPoolProperties` it will be. I will work this in.

@mattrjacobs @lixlix I don't think this is a good default either. If I understand the mechanism correctly, the timer thread pool is for handling timeouts of actions until the circuit opens. If the circuit is open, actions will fail right away and the timer thread pool is no longer used. Therefore, you require a large timer thread pool only if you have a lot of timeouts to deal with (e.g. if circuit breaking is disabled). Furthermore, timeout handling is a *cheap* operation compared to the usual `HystrixCommands` because there is no IO involved (right?). 
I suggest we use a flat default (like 5 or so) and offer a way to configure it. If we want a scaling default, it should scale along the size and count of thread pools used for `HystrixCommands`.

mattrjacobs:
I don't disagree with the flat number of threads approach. I'd like to do some experiments (both using jmh and Netflix production apps) to see what happens when we limit the threads.  

I also wanted to point out that Hystrix collapsing uses the HystrixTimer threads.  Those threads generate ticks and then performing the batching

tine2k:
@mattrjacobs I updated the branch with a `HystrixTimerThreadPoolProperties` implementation. Please review. I left the default as it is. Feel free to change to a flat number after your experiments.

hystrix-core/src/test/java/com/netflix/hystrix/util/HystrixTimerTest.java
```diff
@@ -15,22 +15,35 @@
  */
 package com.netflix.hystrix.util;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
+import com.netflix.hystrix.Hystrix;
```
mattrjacobs: Unused import

hystrix-core/src/main/java/com/netflix/hystrix/strategy/properties/HystrixPropertiesTimerThreadPoolDefault.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.hystrix.strategy.properties;
+
+import com.netflix.hystrix.HystrixThreadPoolKey;
```
mattrjacobs: unused import

hystrix-core/src/main/java/com/netflix/hystrix/strategy/properties/HystrixPropertiesTimerThreadPoolDefault.java
```diff
@@ -0,0 +1,29 @@
...
+import com.netflix.hystrix.HystrixThreadPoolProperties;
```
mattrjacobs: unused import

hystrix-core/src/main/java/com/netflix/hystrix/HystrixTimerThreadPoolProperties.java
```diff
@@ -0,0 +1,79 @@
...
+        public Setter withCoreSize(int value) {
+            this.coreSize = value;
+            return this;
+        }
+
+        /**
+         * Base properties for unit testing.
+         */
+        /* package */
+        static Setter getUnitTestPropertiesBuilder() {
```
mattrjacobs: Does this need to be here?  It appears unused

mattrjacobs:
I left a couple of small cosmetic comments, but this looks great!  Thanks for doing the extra work to get Archaius integration.

mattrjacobs:
Merging.  Will issue the cleanup as a separate commit.  Thanks, @tine2k!

tine2k:
Thanks for your review notes, @mattrjacobs. Sorry, I couldn't work them in in time (I would have). Anyways, thanks for the merge!

Issue #1012, opened 2015-12-07
NullPointerException when pop on Hystrix#list is called Hystrix#endCurrentThreadExecutingCommand


mzeijen:
We sometimes get a warning logged of a NullPointer within the LinkedList of the Hystrix#list field when it's pop methods is called during the Hystrix#endCurrentThreadExecutingCommand execution. It can happen a couple of thousands of times an hour and sometimes almost never

I don't know when this happens and we can't reproduce it. Looks like a multi threading issue where there is some concurrent access on that linked list which causes the problem. 

We see this in Hystrix version 1.4.13 but the Hystrix code regarding this part seems the same in the current 1.4.21 version. Maybe this is already fixed but I can't find anything related in the issues.

Stacktrace:
```
java.lang.NullPointerException
	at java.util.LinkedList.remove(LinkedList.java:791)
	at java.util.LinkedList.removeFirst(LinkedList.java:134)
	at java.util.LinkedList.pop(LinkedList.java:601)
	at com.netflix.hystrix.Hystrix.endCurrentThreadExecutingCommand(Hystrix.java:134)
	at com.netflix.hystrix.Hystrix.access$000(Hystrix.java:32)
	at com.netflix.hystrix.Hystrix$2.call(Hystrix.java:121)
	at com.netflix.hystrix.AbstractCommand.handleThreadEnd(AbstractCommand.java:916)
	at com.netflix.hystrix.AbstractCommand$7.call(AbstractCommand.java:665)
	at rx.Observable$11.onCompleted(Observable.java:4309)
	at rx.internal.operators.OperatorDoOnEach$1.onCompleted(OperatorDoOnEach.java:46)
	at rx.internal.operators.OperatorDoOnEach$1.onCompleted(OperatorDoOnEach.java:53)
	at rx.internal.operators.OperatorOnErrorResumeNextViaFunction$1.onCompleted(OperatorOnErrorResumeNextViaFunction.java:64)
	at rx.internal.operators.OperatorDoOnEach$1.onCompleted(OperatorDoOnEach.java:53)
	at rx.internal.operators.OperatorDoOnEach$1.onCompleted(OperatorDoOnEach.java:53)
	at com.netflix.hystrix.AbstractCommand$HystrixObservableTimeoutOperator$3.onCompleted(AbstractCommand.java:1012)
	at rx.internal.operators.OperatorDoOnEach$1.onCompleted(OperatorDoOnEach.java:53)
[...]

PR #1039, opened 2016-01-07
Make methods of Hystrix class thread-safe

mattrjacobs:
Addresses #1012.  The `testMultipleSemaphoreObservableCommandsInFlight` unit test failed consistently when using a `LinkedList` instead of the `ConcurrentStack`.

Would you mind taking a look, @mzeijen?  Thanks again for bringing this up.  I'll add the deprecation as the next commit if you think this looks like the right approach.

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/Hystrix.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/Hystrix.java
@@ -15,9 +15,9 @@
  */
 package com.netflix.hystrix;
 
-import java.util.LinkedList;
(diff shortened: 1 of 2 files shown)


mattrjacobs:
I'm going to merge this as-is, and open up a new issue for how `Hystrix#getCurrentThreadExecutingCommand()` should handle multiple concurrent semaphore-isolated `HystrixObservableCommand`s, as I think deprecation is 1 solution, but possibly not the only

mzeijen:
Looks great! Thanks :+1:

Issue #943, opened 2015-10-14
Hystrix Metric Streams for Asynchronous Consumers


mattrjacobs:
A different take on #854 is to change the primitive of metrics collection away from the rolling data structures (`HystrixRollingNumber`/`HystrixRollingPercentile`) and into a stream.  

Some consumers of the stream would be asynchronous (`hystrix-servo-metrics-publisher` for example).  Command execution could proceed without any immediate metrics consumption.

It is also possible to imagine consumers that would be synchronous.  These consumers would need the metrics consumption to happen immediately.  Examples of this are `HystrixRequestLog` and `HystrixEventNotifier`.

I propose doing the async cases (and possibly doing the sync cases later) as a way to gain flexibility.  By treating each metrics as an event emission, then consumers of the stream can be arbitrary.  

I'll link a series of Github issues to this issue.

The high-level sequence of work is:
1) Add a metrics stream and write to it
2) Convert current async consumers to use this new stream
3) Remove writes to the rolling data structures.
4) Add new metric consumers as examples of what can be done with a stream-based approach.

mattrjacobs:
Done in #1047

PR #1047, opened 2016-01-12
Hystrix metrics as streams

mattrjacobs:
This (large) set of commits is intended to solve #854 and #943.

The premise of these changes is that the Hystrix library is not the optimal place to decide how metrics should be consumed.  By representing metrics as event streams, consumers gain flexibility in choosing their consumption strategy.

In 1.4.x, all events get aggregated in command-level data structures (`HystrixRollingNumber` and  `HystrixRollingPercentile`).  This is lossy in the sense that only aggregate queries are possible against these data sources.

At the moment, I've achieved feature and performance parity (AFAIK) with 1.4.x while ripping the usage of those data structures out.  The metrics consumers that existed in 1.4.x are now wired against the new metrics streams.  I haven't written new metrics streams that make better use of the new functionality, but that will be coming shortly.

I'm going to start a Wiki page with more details on the 1.5 metrics architecture.  Will link that here when it's ready.  

I've run this code in the Netflix production environment and haven't seen any show-stopping bugs, so this is very likely to become 1.5.0-RC1.  When it does, I'd love feedback on design/usage.  Any sort of code review would be much appreciated, as well.


mattrjacobs:
JMH comparison between 1.4.22 and this PR are here: https://docs.google.com/spreadsheets/d/1a0ERBQJZzlmVqMpuvvdSwbJuXwt0UmkPsFZ97pvgA8o/edit

Highlights are:
* command throughput for both `HystrixCommand` and `HystrixObservableCommand` increased by 10-20%
* In a case where 7 threads are busy executing commands and 1 thread is busy reading metrics as fast as possible (the writeHeavy scenario), command throughput is up by 3-5% and metrics read throughput is up by 200%

Issue #1058, opened 2016-01-20
Distinguish types of metrics into Event/Config/Utilization


mattrjacobs:
I think that there are 3 categories of metrics that Hystrix exposes, and they are mixed together at the moment (up to 1.4.x).

The 3 types are:
* Events (Success, Short-Circuited, Fallback Success, Collapsed, etc).  These make sense in the context of a single request, or aggregated.  They are discrete values that occur at a point in time.
* Config (Command timeout values, Mappings of Commands -> ThreadPools, etc).  These are JVM-scoped, and can be sampled at any point.  These values exist in a given request, but are more accurately viewed as continuous values that change fairly slowly/infrequently
* Utilization (Command concurrency, Thread Pool concurrency, etc).  These are also JVM-scoped like config, but change much more rapidly (as commands enter/leave the system).  They are also continuous, and may be sampled.  They don't fit into a concept of a request naturally.

At the moment, when a metrics publisher collects values, it combines these 3 concepts.  It looks at aggregated event metrics and samples config and utilization values.  I believe that distinguishing these offers more flexibility.  

In an ideal world, I would publish:
* Event metrics on each request completion (including event type and latency) 
* Config metrics sampled very infrequently
* Utilization metrics sampled very frequently

This assumes the existence of a system which can consume a high volume of data and do aggregation off-box.  So it certainly is not for every Hystrix user.  Therefore, I intend to leave the existing metrics pathways as-is, so I don't break any current users of Hystrix.

This issue will track the addition of:
* Request-based event stream
* Config sample stream
* Utilization sample stream

mattrjacobs:
These 3 streams are added in the following form:

* Utilization sample stream, served by SSE and returning text/event-stream JSON (1x per poll interval)
* Configuration sample stream, served by SSE and returning text/event-stream JSON (1x per poll interval)
* Requests stream, served by SSE.  Returns [0..n] requests as text/event-stream JSON (1x per poll interval)

PR #1060, opened 2016-01-22
sample-based SSE streams for configuration/utilization

mattrjacobs:
Addresses part of #1058.  

This PR adds SSE servlets that emit an infinite stream of text data for configuration and utilization.  I also wired them up to the hystrix-examples-webapp


mattrjacobs:
Closed in favor of #1061

PR #1066, opened 2016-01-25
Added PlatformSpecific class that can detect if it's running in AppEngine

### LINKED ISSUE #795: GAE prohibits spawning new threads directly
Issue found after #734 change

mattrjacobs:
If so, provide the AppEngine-specified ThreadFactory.  This should resolve #795.


mattrjacobs:
Merging, as this doesn't change behavior on my local, or on CI.  Still need to confirm it actually works in GAE and will need help doing that

Issue #970, opened 2015-11-12
Hystrix without Archaius


tine2k:
I want to use hystrix (1.4.20) **without** the dependency to archaius. My goal is to not having to deploy the archaius-core.jar with my application as I dont want to deal with its depencies (like commons-lang etc.). As documented in the wiki, this should be possible by implementing a custom `HystrixPropertiesStrategy` and registering it by calling `HystrixPlugins.getInstance().registerPropertiesStrategy()`. Unfortunately, `HystrixPlugins` has a compile depency to archaius (`ConfigurationManager`) in the constructor:
```java
    private HystrixPlugins() {
        try {
            // Load configuration from hystrix-plugins.properties, if that file exists
            ConfigurationManager.loadCascadedPropertiesFromResources("hystrix-plugins");
        } catch (IOException e) {
            // fail silently
        }
    }
```

How I can use my custom `HystrixPropertiesStrategy` without having to deploy archaius-core.jar?

everve:
+1 This is also a blocker for a lot of adoption I have in mind.


On Thu, Nov 12, 2015, at 01:29 AM, Martin Maier-Moessner wrote:
> I want to use hystrix (1.4.20) *without* the dependency to archaius.
> My goal is to not having to deploy the archaius-core.jar with my
> application as I dont want to deal with its depencies (like commons-
> lang etc.). As documented in the wiki, this should be possible by
> implementing a custom HystrixPropertiesStrategy and registering it by
> calling HystrixPlugins.getInstance().registerPropertiesStrategy().
> Unfortunately, HystrixPlugins has a compile depency to archaius
> (ConfigurationManager) in the constructor:


> private HystrixPlugins() {
        try {    // Load configuration from hystrix-plugins.properties,
        if that file exists ConfigurationManager.loadCascadedProperties-
        FromResources("hystrix-plugins"); } catch (IOException e) {
        // fail silently } }
> How I can use my custom HystrixPropertiesStrategy without having to
> deploy archaius-core.jar?


> — Reply to this email directly or view it on GitHub[1].


>


Links:

  1. https://github.com/Netflix/Hystrix/issues/970

mattrjacobs:
Yeah, this will be going into Hystrix 2.0, and is the main reason for that major version bump.  I haven't started work on this yet, but I intend to get to it in the next month or so.

Agreed that this is a pain point.  See #252 for a bit more context

Issue #252, opened 2014-04-29
Archaius version update?


samhendley:
Hystrix is still tied to version 0.4.1 of the archaius library. Can we update to a newer version? I only noticed because that version doesn't correctly expose it's dependencies. In particular it doesn't list commons-lang in its depdencies which gave me an unexpected runtime error (below).

```
[ERROR][scheduler-3][2014-04-29 08:26:01,901] org.springframework.scheduling.support.TaskUtils$LoggingErrorHandler.handleError(TaskUtils.java:95) - Unexpected error occurred in scheduled task.
java.lang.NoClassDefFoundError: org/apache/commons/lang/text/StrLookup
at com.netflix.config.ConfigurationManager.getConfigInstance(ConfigurationManager.java:125)
at com.netflix.config.DynamicPropertyFactory.getInstance(DynamicPropertyFactory.java:263)
at com.netflix.config.DynamicProperty.getInstance(DynamicProperty.java:245)
at com.netflix.config.PropertyWrapper.<init>(PropertyWrapper.java:58)
at com.netflix.hystrix.strategy.properties.HystrixPropertiesChainedArchaiusProperty$DynamicBooleanProperty.<init>(HystrixPropertiesChainedArchaiusProperty.java:300)
at com.netflix.hystrix.HystrixCommandProperties.getProperty(HystrixCommandProperties.java:384)
at com.netflix.hystrix.HystrixCommandProperties.<init>(HystrixCommandProperties.java:116)
at com.netflix.hystrix.HystrixCommandProperties.<init>(HystrixCommandProperties.java:110)
```

mattrjacobs:
From looking at public interfaces which deal with configuration, I believe a 2.0 release will be required.  I'll start outlining the required changes here.

li-a:
[archaius 0.4.1's dependency specifically on commons-configuration 1.8](http://central.maven.org/maven2/com/netflix/archaius/archaius-core/0.4.1/archaius-core-0.4.1.pom) gets it flagged by some security scans because of its [optional dependency on commons-collections 3.2.1](http://central.maven.org/maven2/commons-configuration/commons-configuration/1.8/commons-configuration-1.8.pom), which has [popular security vulnerabilities](https://www.cvedetails.com/version/187982/Apache-Commons-Collections-3.2.1.html).

Seems like this will get resolved only in a 2.x, but there's so little movement on the project lately.. should I stop holding my breath?

Issue #129, opened 2013-03-19
Hystrix without Archaius


daijithegeek:
Hi,

I tried to implement a custom HystrixPropertiesStrategy, based on our company configuration mechanism. I was wondering if it is possible to get rid of the Archaius dependency as I really do not want to use it.

Indeed, any HystrixCommandProperties implementation need to call the super() constructor which enables Archaius class access.

Another option is to implement an configuration datasource for Archaius based on our configuration mechanism. I'm not a big fan of it for different reasons. The main one is the fact that I am currently developing a library, and archaius comes with guava and commons-everything dependencies. Therefore I will force library client to get those dependencies, adding many potential conflict on their side.

Am I missing something important there ? Is there a better way?

Thank you!
David

agentgt:
I would not have a problem with the Archaius dependency if Archaius itself did not depend on so many things that have static initialization (most notably SLF4J and commons-logging).

There really needs to be a boot strap configuration library that does not depend on a logger or various other infrastructure as you get a nice chicken and egg problem of who does the initial configuration. For example you cannot configure your logging infrastructure with Archaius because it will already have instantiated logging. While you can possibly just live with the fact that your logging infrastructure will have to reinitialize this has historically been problematic with libraries such as logback (in fact there was no way to control the initialization of logback till [I filed a bug](https://github.com/qos-ch/logback/pull/181)).

One of the ways to mitigate dispatching to an implementation with out relying on system properties or environment variables is to use the JDK Service Loader (which is what I did for logback). Consequently depending on what jars are in the classpath the right plugins would then be loaded automatically.

TLDR ... depending on libraries that have any sort of dependency of static initialization is hard to remove and makes it difficult for those who care about initialization (usually logging is the primary issue).

mattrjacobs:
@agentgt I agree 100%.  I'll be working on this soon

agentgt:
For future folks that land here see #970

PR #1083, opened 2016-02-02
Make Archaius a soft dependency through reflection and improve plugin…

agentgt:
See #970 #129 #252 

This change does a couple of things to `hystrix-core` that should be fairly safe and backward compatibility
 
 * It checks to see if Archaius is available through reflection and will use it if is available. Notice I did not change the dependencies so backward compatibility should be fine.
 * Besides using properties for plugin discovery the [JDK service loader](https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html) is used as a fallback.
 * I added a new plugin called `HystrixDynamicProperties` due to the whole nasty coupling with `HystrixPropertiesChainedArchaiusProperty` and `PropertyWrapper`. Thus you don't have to replace every `XXXStrategy` if you only want to change configuration source.
 * The `HystrixDynamicProperties` can only be loaded through a `ServiceLoader` (cause you got not properties to resolve it :) ). Thus `HystrixPlugin` must load with some plugin implementation and thus is `final`. If no implementation is found an implementation that uses the `System.properties` is used ( we could instead fail). This would only happen if you didn't have Archaius in your classpath and no other implementation available (ie serviceloader).
 * There are a whole bunch of contrib stuff that uses Archaius directly that I did not fix but in theory could be.

I need these changes because right now my company cannot use Archaius and thus we have our own nasty circruitbreaker stuff that I would love to get rid of.

I know @mattrjacobs was working on this as well. This is mainly to invoke discussion. I'm also not so sure what the coding style is or should be (I saw tabs but used spaces...). Advance apologies.


mattrjacobs:
Thanks for the PR, @agentgt .  Very good timing, as I was just working on this.  Will review and leave comments.

agentgt:
Thanks @mattrjacobs . I can cleanup the code and add Javadoc tomorrow. The only downside to this implementation is that if you really really do not want Archaius you will have to manually exclude the dependency (since if we make the dependency optional it will break existing users). But even if it is in the classpath one could provide another `HystrixDynamicProperties` implementation through the ServiceLoader mechanism and thus prevent loading (since Java lazy loads classes).

This is also fairly hard to unit test with out creating dummy projects but I have done something similar in the past to test something like this (apologies for no unit test). I'll see if I can add a unit test later.

mattrjacobs:
I took a first pass through this and it looks like it does what I think it should.  I will set up a couple projects in hystrix-examples to get some more experience with it before I merge.  

Another problem that I'd like to solve is the actual compile dependency on Archaius.  I'd like to drop that and make a new contrib module `hystrix-archauis` that provides that binding.  I know that's a bit of work - I was already planning on doing it, so I'm happy to either do that myself or you can.  

Once I get some examples, with and without Archaius, working, I'll layer those on top of this PR.

Thanks for this!

agentgt:
@mattrjacobs I thought about making a `hystrix-archaius` module but I couldn't figure out a way with out seriously breaking most of the unit tests and/or having a cyclic dependency since the `hystrix-core` unit tests would need `hystrix-archaius` and the `hystrix-archaius` module would need  `hystrix-core` for the interfaces.

The other reason I didn't do it was that I was hoping not to have to wait till version 2.0 for this to be released. @benjchristensen had alluded that making sweeping changes in the config loading (and existing interfaces in general) would have to wait till 2.0 (something my company can't really wait for albeit we can just for fork for now). I'm not sure if the same applies for unit tests but I rather be safe and have a more incremental process (ie release more).

Speaking of which I'll be adding unit tests shortly to test the ServiceLoader part :) (the reflection is already heavily hit with the existing tests).

agentgt:
@mattrjacobs I have finished adding Unit tests and Javadoc. The only thing I noticed/changed is how the HystrixPlugin is loaded as a singleton eagerly.

```java
    //We should not load unless we are requested to. This avoids accidental initialization. @agentgt
    //See https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom
    private static class LazyHolder { private static final HystrixPlugins INSTANCE = HystrixPlugins.create(); }
```

I prefer lazy loading so you don't accidentally kick off loading by just importing `HystrixPlugin`. There is a slight chance this could affect backward compatibility and or performance (but this is very dubious). This may have been even done on purpose. Perhaps @benjchristensen would know?

Let me know if and when you want me to squash the commits.

mattrjacobs:
This change is already backwards incompatible with 1.4/1.5, due to the type hierarchy change.  When I swapped out hystrix-core-1.4.23.jar with a jar built from this branch, I got a `java.lang.VerifyError` that included this:

```
Reason:
    Type 'com/netflix/hystrix/strategy/properties/HystrixPropertiesChainedArchaiusProperty$DynamicBooleanProperty' (current frame, stack[3]) is not assignable to 'com/netflix/config/PropertyWrapper'
```

Changing the type hierarchy is not permissible.  So it's already looking like this might have to go to 2.0.

agentgt:
@mattrjacobs If that is the case we should probably rename that class since its no longer tied to Archaius and make the other classes in there (`HystrixPropertiesChainedArchaiusProperty`) final and maybe even protected/default.

For now my company can just roll and release our own in our private repo.

agentgt:
@mattrjacobs I think I can fix the hierarchy change easily (well somewhat) by just creating another `HystrixPropertiesChainedArchaiusProperty` and calling it something else probably `HystrixPropertiesChainedProperty` and then just update all the default strategies to that class. That is I would just leave `HystrixPropertiesChainedArchaiusProperty` alone.

I'll update the PR with this change shortly.

agentgt:
@mattrjacobs I tested the latest in isolation without Archaius in the classpath and it now seems to work fine. When you get a chance can you test on your end as I can't test the verify hierarchy  issue you had (n.b. I accidentally amended some commits and force pushed).

The only change I want to make is perhaps simplify `HystrixPropertiesChainedProperty` (which replaces the existing archaius one but the archaius one remains for existing client code). It just has too many `public` code touch points and there doesn't appear to be an `internal` package for Hystrix core. For 2.0 there probably should be so we don't have these backward compatibility issues.

mattrjacobs:
@agentgt I talked to @elandau (owner of Archaius) and got some more clarity on where things stand.

Let me know if the following plans make sense for you:

1) Keep going with this PR.  If we can ensure backwards-compatibility, then it can get released in 1.5.0.  The functional change would be that Hystrix can work without Archaius on the classpath.  This should solve your immediate need.  As a side note, I'm still not sure about `j.u.ServiceLoader` - I need to do more research/experimentation to understand it better.

`hystrix-core` would still depend on Archauis, as some pieces of the inheritance hierarchy are in Archaius packages (`com.netflix.config`).  However, we'd start establishing the path where configuration implementation is decoupled.

2) In Hystrix 2.0, remove Archaius from `hystrix-core`.  Configuration objects would still exist, but the only implementation in `hystrix-core` would be one with no external library dependencies - something like `j.u.ServiceLoader` or some other strategy that pulls config from properties files at app start

Then, add in `hystrix-archauis` and `hystrix-archaius2` submodules that can fulfill the config interfaces in `hystrix-core`.  Archaius 2.0 was just released, and has a much tighter dependency set.  If other configuration mechanisms are useful, submodules to integrate them could be introduced as well.

Let me know if this makes sense.

agentgt:
Thanks @mattrjacobs for the followup!

As for the ServiceLoader it does have some rather undeserved criticism  on some stackoverflow post (that I can't seem to find) but the fact is its used heavily in the Java world during bootstrapping and thus the class is generally already loaded. In fact the logging library logback uses the ServiceLoader (I would know because I put it in :smiley: ) .

You could certainly load using a System property but I believe this causes more issues and is not canonical. The service loader also has the advance benefit if you just put a jar with the right META-INF registration it will load it.. ie plugins on demand based on dependencies.

As for Archaius 2.0 I really hope it does well but I found it still to be rather buggy (I'm in fact surprised there is now a 2.0 tag... and it has been released). 

I'm really hoping Archaius becomes the slf4j for configuration. I wrote my own [configfacade](https://github.com/agentgt/configfacade) lib for my company because there are so many libs/frameworks now that used a mixture of configuration formats/sources and I need an abstraction.

When we figure out the 1.5 stuff I'll continue to help port the code and/or fix all the unit tests that reference Archaius (directly or indirectly).

mattrjacobs:
@agentgt I've just confirmed that this version is binary-compatible with 1.4.x.  Thanks for getting that done!  I'll take another pass through your changes to see if anything else requires review.  

I've only tested the case in which Archaius is provided, and verified that this still works.  Have you tested that excluding Archaius drops back to static configuration?

agentgt:
@mattrjacobs Yes I have manually tested with Archaius not in the classpath and it works (it uses the system properties).

I'm thinking of checking a system property to disable loading archaius even if it is on the classpath so that we can at least have unit test coverage for that execution path of falling back to system properties (I already have a test for the service loader path).

That should be my last change.

agentgt:
@mattrjacobs I added some more unit tests and way to load the HystrixDynamicProperties via a system property (`"hystrix.plugin.HystrixDynamicProperties.implementation"`).

This made unit testing a little easier as well as providing an option to completely avoid the service loader for those that might be concerned about the performance of initial resource checking (ie cmdline
`-Dhystrix.plugin.HystrixDynamicProperties.implementation=someclass`).

I don't think I have any more changes. Thank you!

mattrjacobs:
Thanks, @agentgt .  Reviewing now...

hystrix-core/src/main/java/com/netflix/hystrix/strategy/properties/archaius/HystrixDynamicPropertiesArchaius.java
```diff
@@ -0,0 +1,116 @@
...
+package com.netflix.hystrix.strategy.properties.archaius;
+
+import com.netflix.config.PropertyWrapper;
+import com.netflix.hystrix.strategy.properties.HystrixDynamicProperties;
+import com.netflix.hystrix.strategy.properties.HystrixDynamicProperty;
+
+/**
+ * This class should not be imported from any class in core or else Archaius will be loaded.
+ * @author agentgt
+ * @see HystrixArchaiusHelper
```
mattrjacobs: This doesn't resolve.  Can you add the {@link XXX} syntax and either import `HystrixArchaiusHelper` or use the fully-qualified class name?

mattrjacobs:
I'm going to try to summarize the changes to make sure I understand them.  Please correct me if I got any of the details wrong.

* The first time `HystrixPlugins` gets referenced, nothing happens, due to use of `HystrixPlugins.LazyHolder`
* The first time `HystrixPlugins.getInstance()` gets referenced, that kicks off the work to create an instance.  This should happen either via a user calling `HystrixPlugins.getInstance().registerXXX()` to set up something environment-specific, or as the first command executes, when it tries to start resolving plugin instances
* The creation of the `HystrixPlugins` instance sets up the instance of `HystrixDynamicProperties`, by checking the following, in order:
  * Using System.properties to resolve "hystrix.plugin.HystrixDynamicProperties.implementation"
  * Using `ServiceLoader` to find an instance of `HystrixDynamicProperties`.
  * If `com.netflix.config.ConfigurationManager` is on the classpath:
    * Look in resource `hystrix-plugins` and use Archaius to load it
    * Create `com.netflix.hystrix.strategy.properties.archaius.HystrixDynamicPropertiesArchaius`
  * Use System.properties

Because of that loading algorithm and the use of lazy loading, Archaius is not referenced until you've had a chance to override it.  And not finding it still allows the System properties to work as an implementation of `HystrixDynamicProperties`.  That instance (however it is arrived at), is used in all other plugin lookups.

For command/threadpool/collapser/timer-threadpool properties, these are all instances of `HystrixPropertiesChainedProperty`.  As a result, they can be chained and support an API that isn't tied to Archaius (nice work here!).  Invocations of `add()` result in calls to `HystrixPropertiesChainedProperty.getDynamicProperty()`, which completes the loop by calling back to `HystrixPlugins.getInstance().getDynamicProperties()`.  That was what got set up by the bullet-points above.

This all makes sense to me, and I think is a valuable addition.  I left a couple of line-level comments as well, but wanted to make sure I reviewed the changeset as a whole somewhere (and that I understand how it hangs together).

Thanks for the contribution!

hystrix-core/src/main/java/com/netflix/hystrix/strategy/HystrixPlugins.java
```diff
@@ -295,6 +345,32 @@ private static Object getPluginImplementationViaArchaius(Class<?> pluginClass) {
             return null;
         }
     }
+    
+    
+
+    private static HystrixDynamicProperties resolveDynamicProperties(ClassLoader classLoader) {
```
mattrjacobs: IMO, I think it's valuable to add a logger.info call here on each possible return.  This will aid debugging as users can understand where the `HystrixDynamicProperties` instance is getting loaded from.

Something like : 
`logger.info("Created HystrixDynamicProperties instance from System property named \"hystrix.plugin.HystrixDynamicProperties.implementation\");

or

logger.info("Created HystrixDynamicProperties instance by loading from ServiceLoader.  Using class : " + hp.getClass().getSimpleName()");

As examples of verbosity.  

What do you think?

hystrix-core/src/main/java/com/netflix/hystrix/strategy/HystrixArchaiusHelper.java
```diff
@@ -0,0 +1,90 @@
...
+     * @ExcludeFromJavadoc
+     */
+    static boolean isArchaiusV1Available() {
+        return LazyHolder.loadCascadedPropertiesFromResources != null;
+    }
+
+    static void loadCascadedPropertiesFromResources(String name) {
+        if (isArchaiusV1Available()) {
+            try {
+                LazyHolder.loadCascadedPropertiesFromResources.invoke(null, name);
```
mattrjacobs: What about the return value here?  If we find implementations in the `hystrix-plugins` file, shouldn't we try to instantiate them and return them.  Or am I missing something?

hystrix-core/src/main/java/com/netflix/hystrix/strategy/HystrixPlugins.java
```diff
@@ -36,32 +37,54 @@
...
-            // Load configuration from hystrix-plugins.properties, if that file exists
-            ConfigurationManager.loadCascadedPropertiesFromResources("hystrix-plugins");
-        } catch (IOException e) {
-            // fail silently
-        }
+    
+    private HystrixPlugins(ClassLoader classLoader) {
+        //This will load Archaius if its in the classpath.
+        this.classLoader = classLoader;
+        //N.B. Do not use a logger before this is loaded as it will most likely load the configuration system.
```
agentgt: Load config before loading logging framework.

hystrix-core/src/main/java/com/netflix/hystrix/strategy/HystrixPlugins.java
agentgt: See my other line comment on logging. We have to be very careful about when we kick off logging. It needs to be after whatever configuration framework has loaded (`HystrixDynamicProperties`) because the configuration framework may want to participate in configuring logging. This is a long going chicken/egg problem with java config and logging. This appears to be after so we can probably do it **but we cannot statically call `getLogger` or else we load the logging framework too early.**  I'll see if I can put in a safe change.

hystrix-core/src/main/java/com/netflix/hystrix/strategy/properties/archaius/HystrixDynamicPropertiesArchaius.java
agentgt: I will remove the line altogether. HystrixArchaiusHelper  now is private and internal anyway (hence why I moved it to the same package as `HystrixPlugin`).

hystrix-core/src/main/java/com/netflix/hystrix/strategy/HystrixArchaiusHelper.java
agentgt: Basically this code comes from the original 1.4 code https://github.com/Netflix/Hystrix/blob/1.4.x/hystrix-core/src/main/java/com/netflix/hystrix/strategy/HystrixPlugins.java#L57

What it does is indirectly boots up Archaius and adds the hystrix-plugins properties file as an additional configuration resource. The actual method does not return anything.The assumption here is Archaius isn't going to provide configuration properties to point to use a different `HystrixDynamicProperties`.

This is only to create `HystrixDynamicProperties`... all the other implementations will be loaded using cascaded configuration (assuming you have Archaius). That is `hystrix-plugins.properties` will be used.

This does bring up an excellent point though that the `HystrixDynamicPropertiesSystemProperties` (I picked that name to be consistent with the naming scheme) perhaps should load `hystrix-plugins.properties`? Let me know what you think about that.

agentgt:
@mattrjacobs I added logging (safely) and fixed the javadoc issue. Let me know what you think.

mattrjacobs:
Any reason for mixing debug/info level logging?  My preference would be for all to be at info.

mattrjacobs:
I just tested this again in an Archaius environment, and that worked as expected.

agentgt:
> Any reason for mixing debug/info level logging? My preference would be for all to be at info.

Backward compatibility. Existing users might not like the noise... ie we are adding an info statement that didn't happen before.

mattrjacobs:
OK, works for me.  Thanks for all your work getting this done.  I'll merge and cut RC3 shortly.  I'd love any  validation you could do on that.

mattrjacobs:
Released in 1.5.0-rc.3 now.

Issue #995, opened 2015-11-26
@CacheResult on @HystrixCommand causes a NPE


AlexejK:
Hi,

I'm hitting a bit of a confusing issue when trying to use request caching with Hystrix Command.
The `@HystrixCommand` without `@CacheResult` annotation works perfectly fine, however as soon as I add the annotation, it fails with a NullPointerException.

```
Caused by: java.lang.NullPointerException: null
	at com.netflix.hystrix.HystrixRequestCache.get(HystrixRequestCache.java:104) ~[hystrix-core-1.4.20.jar:1.4.20]
	at com.netflix.hystrix.AbstractCommand.toObservable(AbstractCommand.java:346) ~[hystrix-core-1.4.20.jar:1.4.20]
	at com.netflix.hystrix.HystrixCommand.toObservable(HystrixCommand.java:42) ~[hystrix-core-1.4.20.jar:1.4.20]
	at com.netflix.hystrix.HystrixCommand.queue(HystrixCommand.java:378) ~[hystrix-core-1.4.20.jar:1.4.20]
	at com.netflix.hystrix.HystrixCommand.execute(HystrixCommand.java:335) ~[hystrix-core-1.4.20.jar:1.4.20]
```

The HystrixCommand takes one argument, which is a string, CacheKey is being detected properly but for some reason I'm always getting this NPE.

dmgcodevil:
will try with last javanica version. 1.4.20 still uses bogus ObservableResult

AlexejK:
Well this was my fault to begin with: I did not really initialise the `HystrixRequestContext` as described here: https://netflix.github.io/Hystrix/javadoc/com/netflix/hystrix/strategy/concurrency/HystrixRequestContext.html
I Assume this is still needed even when using Javanica.

However, NPE should not be happening. In Hystrix 1.4.18 I was actually able to get an exception clearly stating that `HystrixRequestContext.initializeContext()`has to be called first.

dmgcodevil:
yes, context must be initialized as with venila hystrix.

AlexejK:
So I guess we can conclude that there is simply a problem of an NPE being cast instead of the exception that was given pre 1.4.20 that notified about `HystrixRequestContext.initializeContext()`.

PR #1093, opened 2016-02-06
Fixed NPE on request cache when HystrixRequestContext is not initialized

caarlos0:
I'm new to Hystrix, so I'm not sure this is the ideal implementation.

This might fix https://github.com/Netflix/Hystrix/issues/995 .

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/HystrixRequestCache.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/HystrixRequestCache.java
@@ -102,7 +102,7 @@ public class HystrixRequestCache {
         ValueCacheKey key = getRequestCacheKey(cacheKey);
         if (key != null) {
             if (!HystrixRequestContext.isCurrentThreadInitialized()) {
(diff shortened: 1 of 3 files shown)


caarlos0:
Apparently there are unstable tests..

image

It also passed in Jenkins and travis... but int the last build, for some reason, these two tests failed:

- com.netflix.hystrix.HystrixCommandTest#testSemaphorePermitsInUse
- com.netflix.hystrix.HystrixCommandTest#testRequestCacheWithSlowExecution

Any tips?

mattrjacobs:
@caarlos0 Thanks for the contribution - looks like a good change to me.  One subtlety here is that the exception thrown is synchronous, while the method is returning an `Observable<T>`.  I think that this is appropriate, as it distinguishes an actual error in the command data path (async `onError`) versus an error in the command setup.

I'll merge this and look at the flaky unit tests.

caarlos0:
@mattrjacobs Nice to hear, thanks for the great lib, btw :beers:

PR #1105, opened 2016-02-25
Allow Hystrix  to be used as lambda

velo:
The end game here is to be able to use Hystrix with java 8 lambdas
```
    try (HystrixRequestContext context = HystrixRequestContext.initializeContext();)
    {
      int result = execute(asKey("unittest"), () -> number * 10);
      Assert.assertEquals(70, result);
    }
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: java
-sudo: false
+sudo: required
+dist: trusty
 
 jdk:
   - oraclejdk7
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: java
-sudo: required
-dist: trusty
+sudo: false
 
 jdk:
   - oraclejdk7
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Hystrix Releases #
 
+### Version 1.5.0-rc.4 ([Maven Central](http://search.maven.org/#search%7Cga%7C1%7Cg%3A%22com.netflix.hystrix%22%20AND%20v%3A%221.5.0-rc.4%22), [Bintray](https://bintray.com/netflixoss/maven/Hystrix/1.5.0-rc.4/)) ###
+
+This version does not have any known bugs, but is not recommended for production use until 1.5.0.
+
+Included changes: 
+
+* [Pull 1099](https://github.com/Netflix/Hystrix/pull/1099) Bugfix to get Hystrix dashboard operational again
+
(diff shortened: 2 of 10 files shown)


mattrjacobs:
My initial thought is that this is a fine thing to do, but belongs outside of `hystrix-core`.  This change essentially separates out `HystrixCommand` into 3 pieces: GroupKey, execution behavior, and fallback behavior.  

There's much more that's possible to configure on a `HystrixCommand`, and I think matching the config capabilities of a direct `HystrixCommand` object takes away from the readability gains of the lambda-syntax.  

`hystrix-core` also has functionality which you haven't modeled yet: `HystrixObservableCommand`, and the `queue()`,`observe()`, `toObservable()` method of executing commands.  It's completely fine to not model those pieces of functionality if your use case doesn't require them, but then this code should be outside of `hystrix-core`.

Would anyone else in the community use this feature if added?

mattrjacobs:
Also, I think adding the Closeable interface makes sense, regardless of the rest of the PR.  If you'd like to submit that as a standalone PR, I'd be happy to merge that in.

velo:
Done...
https://github.com/Netflix/Hystrix/pull/1111

velo:
@mattrjacobs I prefer to include `queue()`, `observe()`, `toObservable()` =D

So, how does it feel to be able to do this:
```
    try (HystrixRequestContext context = initializeContext();) {
      int result = queue(asKey("unittest"), () -> number / 0, () -> -1)
          .get(1, TimeUnit.SECONDS);
```

observable:
```
    try (HystrixRequestContext context = initializeContext();) {
      observe(asKey("unittest"), () -> number / 0)
        .map(arg -> arg.byteValue())
        .toBlocking()
        .forEach(asByte -> System.out.println(asByte));
    }
```

mattrjacobs:
How do you intend to pass constructor arguments to command instances?

velo:
Overloading with HystrixCommand.setter should cover all constructor arguments right?
```
    try (HystrixRequestContext context = initializeContext();) {
      int result = queue(
          Setter
              .withGroupKey(asKey("unittest"))
              .andThreadPoolKey(HystrixThreadPoolKey.Factory.asKey("testpool")),
          () -> number / 0,
          () -> -1)
              .get(1, TimeUnit.SECONDS);
```

So far I have only used `HystrixCommandGroupKey`, shame on me

mattrjacobs:
That's one aspect of this - configuration of commands is sometime fairly detailed, sometimes not.

The other aspect is that almost all useful concrete impls of HystrixCommand accept domain arguments (customer ID, country, widget ID, etc...)

So straightline code without your syntactic sugar looks something like:

```
HystrixCommand<Widget> cmd = new WidgetCommand(12345L, Country.US)
Observable<Widget> w = cmd.observe();
```

In this case assume that `run()` / `getFallback()` are implemented in a standard way.

So in your lambda case, your closures for `run()` / `getFallback()` would be able to use variables from the current scope.  So, I don't think there's a loss of functionality.

I'm still on the fence about whether to include this proposal or not.  Is anyone on the sidelines willing to chime in and say whether you think that would be a useful addition?

mattrjacobs:
Closing due to inactivity.  Please re-open if there's more to discuss.

PR #1108, opened 2016-02-26
Javanica HystrixRequestCacheManager should use the concurrency strategy from HystrixPlugins

jhinch:
This can result in unexpected results if the concurrency strategy has overridden getRequestVariable. In my case this resulted in a NPE when clearing the cache entry


mattrjacobs:
Thanks for the submission @jhinch .  Will do a pass over changes to master, see if anything else should get backported to 1.4.x, and do a release later today

jhinch:
Thanks!

PR #1111, opened 2016-02-29
Made HystrixRequestContext Closeable

velo:
now we can leverage try-with-resources on java 7 and forward


mattrjacobs:
Thanks @velo !

Issue #1109, opened 2016-02-29
Make queue size of MetricJsonListener configurable


anuragw:
Currently, the MetricJsonListener in the hystrix-metrics-event-stream collects metrics into a queue which is hard-coded to a size of 1000. Our system has a very high request rate, hence this queue gets populated quite soon when pulling the metrics stream. While we're using version 1.4.23 right now, the latest RC also has the same issue.

Could this queue size be made configurable, perhaps with a default set to the current value of 1000, in order to allow more flexibility for larger systems? Also, if this is agreeable, and the change is included in the current/next release candidate, what would be an approximate ETA on the next release?

mattrjacobs:
Yeah, this sounds like an OK thing to do.  I'm happy to review a PR for this.

anuragw:
My upstream git config got messed up, I'll probably have to rewrite the history and force-push. I'll let you know once its done, there should only be the original two commits I had.

mattrjacobs:
Closed via #1114

anuragw:
Thanks, @mattrjacobs!

anuragw:
@mattrjacobs Any updates on when 1.4.25 might be released? I'm assuming this change goes in the 1.4.25 release.

mattrjacobs:
Just released in 1.4.25: https://github.com/Netflix/Hystrix/releases/tag/v1.4.25

Haven't done a release out of master yet - I need to evaluate a couple other changes/proposals before I cut that.

PR #1114, opened 2016-03-02
#1109 Make queue size of MetricJsonListener configurable

anuragw:
The queue size of the MetricJsonListener used by the HystrixMetricsStreamServlet has been made configurable in two ways:

- new property "hystrix.stream.defaultMetricListenerQueueSize", which if unspecified will be set to 1000
- optional request parameter "queueSize", which can override the default queue size for each request

The changes have been made on the 1.4.x, but can also be propagated to the 1.5.x RC branch. Please let me know if you need any further information or would like to suggest changes.

### DIFF
--- a/hystrix-contrib/hystrix-metrics-event-stream/src/main/java/com/netflix/hystrix/contrib/metrics/eventstream/HystrixMetricsStreamServlet.java
+++ b/hystrix-contrib/hystrix-metrics-event-stream/src/main/java/com/netflix/hystrix/contrib/metrics/eventstream/HystrixMetricsStreamServlet.java
@@ -135,7 +135,17 @@ public class HystrixMetricsStreamServlet extends HttpServlet {
                 response.setHeader("Cache-Control", "no-cache, no-store, max-age=0, must-revalidate");
                 response.setHeader("Pragma", "no-cache");
 
-                MetricJsonListener jsonListener = new MetricJsonListener();
+                int queueSize = 1000;
+                try {
+                    String q = request.getParameter("queueSize");
+                    if (q != null) {
+                        queueSize = Math.max(Integer.parseInt(q), queueSize);
+                    }
+                } catch (Exception e) {
+                    // ignore if it's not a number
+                }
+
+                MetricJsonListener jsonListener = new MetricJsonListener(queueSize);
                 poller = new HystrixMetricsPoller(jsonListener, delay);
                 // start polling and it will write directly to the output stream
                 poller.start();
@@ -207,7 +217,11 @@ public class HystrixMetricsStreamServlet extends HttpServlet {
          * <p>
          * This is a safety check against a runaway poller causing memory leaks.
          */
-        private final LinkedBlockingQueue<String> jsonMetrics = new LinkedBlockingQueue<String>(1000);
+        private LinkedBlockingQueue<String> jsonMetrics;
+
+        public MetricJsonListener(int queueSize) {
+            jsonMetrics = new LinkedBlockingQueue<String>(queueSize);
+        }
 
         /**
          * Store JSON messages in a queue.
--- a/hystrix-contrib/hystrix-metrics-event-stream/src/main/java/com/netflix/hystrix/contrib/metrics/eventstream/HystrixMetricsStreamServlet.java
+++ b/hystrix-contrib/hystrix-metrics-event-stream/src/main/java/com/netflix/hystrix/contrib/metrics/eventstream/HystrixMetricsStreamServlet.java
@@ -15,22 +15,20 @@
  */
 package com.netflix.hystrix.contrib.metrics.eventstream;
 
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.concurrent.LinkedBlockingQueue;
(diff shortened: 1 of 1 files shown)


hystrix-contrib/hystrix-metrics-event-stream/src/main/java/com/netflix/hystrix/contrib/metrics/eventstream/HystrixMetricsStreamServlet.java
```diff
@@ -135,7 +134,17 @@ private void handleRequest(HttpServletRequest request, HttpServletResponse respo
                 response.setHeader("Cache-Control", "no-cache, no-store, max-age=0, must-revalidate");
                 response.setHeader("Pragma", "no-cache");
 
-                MetricJsonListener jsonListener = new MetricJsonListener();
+                int queueSize = defaultMetricListenerQueueSize.get();
+                try {
+                    String q = request.getParameter("queueSize");
```
spencergibb: Why would you let the request set the queue size?

anuragw: If a system talks to multiple, dynamically scaleable endpoints, it could be cumbersome to change the queue size only via a property. Providing a request parameter would allow the hystrix-dashboard or other visualization service to be reconfigured in a much easier way.

mattrjacobs: IMO, the velocity of metrics is independent of who is consuming the metrics.  What problem do you intend to solve by allowing the request to set the queue size?

anuragw: As I mention in my previous comment, there doesn't seem to be a way of scaling the queue size if the number of backend endpoints is scaled up or down.

For instance, a user-facing email service may only need 25 instances of a backend email retrieval service during off-peak but auto-scale to 100-150 during peak periods. In such a case, we would need to estimate the peak load, and set the queue size accordingly. If the volume of metrics exceeded that, the hystrix stream servlet would need to be reconfigured each time.

The consumer of metrics would be in a better position to anticipate the typical patterns in load, and adjust accordingly if it noticed the stats collection to be failing due to hitting queue size caps.

mattrjacobs:
@anuragw I don't follow your reasoning.

The `HystrixMetricsPoller` gets invoked within an infinite loop in `HystrixMetricsStreamServlet`.  It loops over all of the metrics and gets the current value of the metric.  There's a delay parameter in the URL that determines how frequently to generate these metrics (default is 500ms).  Call this value D.

The `HystrixMetricsPoller` produces 1 String for each command + threadpool + collapser in your running system.  Call that value M.  So the rate at which the queue fills up is M/D.

The metrics consumption happens in the servlet thread and also has the same delay.  It reads all values in the queue.

So the `delay` parameter to the servlet governs how quickly to both produce/consume metrics.  But the number of metrics in the system doesn't really change over time, unless you're adding commands/threadpools/collapsers dynamically.

anuragw:
@mattrjacobs your last line says it all, "unless you're adding commands/threadpools/collapsers dynamically", our system does have this behavior.

One other option might be to configure the queueSize property to 100K or 1M, something huge, then use a very tiny delay, say 1ms. But wouldn't this tiny delay be very close to normal processing overheads in the servlet, making it hard to interpret the reported metrics?

anuragw:
I've fixed the issue I had on my forked branch, and have pushed again. The PR got auto-closed when I reset to the base fork before replaying my two commits, and so I've reopened it. I'm working on the build failure now and will update once that's done.

Can we make a call on whether or not to include the queueSize in the request parameters soon? I wouldn't mind keeping the property as the sole way of setting the queueSize, but I do think the request parameter could be useful too, in some scenarios.

anuragw:
I'm not sure what the issue is, the gradle build seems to pass on my system. Any ideas why this may be failing?

mattrjacobs:
Even in the case where commands are being added dynamically, there's still a single set for the JVM.  So I still don't see how defining a queueSize per request helps you.

I'm trying to clean up a bunch of the unit tests to get rid of the flaky failures of unit tests.  They almost always have to do with Travis running things more slowly than our local machines.  I hope to have that significantly cleaned up in the next couple of days

anuragw:
@mattrjacobs Let's agree to disagree on whether or not including the queueSize in the request makes sense. I'll remove the request parameter logic and push out the new code today. Once the unit tests are fixed, is it safe to assume that the PR merge shouldn't take much longer?

mattrjacobs:
Thanks for the good discussion, and contribution, @anuragw.  I plan on getting a release with this change out early next week

anuragw:
@mattrjacobs, I'm always looking forward to contributing! Thanks for your help too, and I'll keep an eye out for the next 1.4.x release.

PR #1116, opened 2016-03-03
Fix underscore.js URL

exys666:
Fix underscore.js URL to work correctly both on http:// and https://
hystrix


mattrjacobs:
Thank @exys666 !

Issue #1033, opened 2015-12-31
Roadmap/direction of hystrix-dashboard


mattrjacobs:
Internally at Netflix, we've stopped using the Hystrix dashboard.  We've been using other visualization approaches, as described here: http://techblog.netflix.com/2014/01/improving-netflixs-operational.html.

We recognize that the dashboard is still relied on by many organizations and individuals, so we'd like to see it continue.  However, it doesn't make sense for me personally, or Netflix as a whole, to be driving its development anymore.

My current thinking is as follows:
- When possible (maybe ~1 month), split off the dashboard into a separate repo.  In this repo, document the fact that feature development is no longer ongoing by me or anyone at Netflix.
- I'll keep admin rights so that I may merge feature requests/bugfixes by others.
- If someone would like to take over as maintainer of the project, I can grant admin rights and then help get all of the documentation/procedures in place so that I can handoff the project and roadmap.

I'd love input on all of the above, and am happy to answer any questions as well.

the4yeast:
@mattrjacobs has Netflix open-sourced any of the tools they used to replace the dashboard?

mattrjacobs:
@the4yeast As of today, no.  That may change as the tools get more mature.

the4yeast:
Thanks @mattrjacobs

PR #1147, opened 2016-03-16
Proposal: HystrixRequestContext junit rule

caarlos0:
This will make testing commands that requires a `HystrixRequestContext`
easier in JUnit by using its Rule feature.

An example of usage would be something like:

```java
    @Rule
    public HystrixRequestContextRule request = new HystrixRequestContextRule();

    @Test
    public void blah() {
       // exec a HystrixCommand that requires some kind of caching, for example
    }
```

I put it in the `hystrix-contrib` folder, let me know if this is the right thing to do, please.

### DIFF
--- a/hystrix-contrib/hystrix-junit/build.gradle
+++ b/hystrix-contrib/hystrix-junit/build.gradle
@@ -1,0 +1,4 @@
+dependencies {
+  compile project(':hystrix-core')
+  compile "junit:junit:4.11"
+}
--- a/hystrix-contrib/hystrix-junit/src/main/java/com/hystrix/junit/HystrixRequestContextRule.java
+++ b/hystrix-contrib/hystrix-junit/src/main/java/com/hystrix/junit/HystrixRequestContextRule.java
@@ -1,0 +1,52 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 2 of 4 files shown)


caarlos0:
(build seems unstable)

spencergibb:
:+1:

mattrjacobs:
Thanks for the contribution, @caarlos0 .  I'm working on increasingly the stability of the CI jobs currently.

This seems like a fine change, but I'd like to hold off for a couple of days.  I'm trying to figure out how to do things like  #1033, where I separate out the lifecycle of hystrix-dashboard from hystrix-core.  That potentially has an impact on hystrix-contrib, depending on the details, so I'd like to get that clear in my head before I take on new items in hystrix-contrib.

I will try to have a better, more concrete answer by next week for you.  If I don't, please pester me until I do :)

caarlos0:
@mattrjacobs Nice! Will do, thanks!

mattrjacobs:
@caarlos0 Here's my thought process:

I would like to only include modules in Netflix/Hystrix that Netflix actually uses.  In practice, I don't have the bandwidth to do more than bugfixes on anything else.  I also become a bottleneck for modules that don't fall into that category.  To be clear, this isn't precisely where we're at today, but it's the direction I'd like to head.

With that in mind, this does seem like something that we would use internally (and it's nice and small), so this seems fine to include in hystrix-contrib.  Thanks for the contribution (and for the patience)!

caarlos0:
@mattrjacobs Nice, thanks =D

I can help to migrate tests to use it, too =)

Issue #1150, opened 2016-03-16
private method getProperty is unused


fcalderonnearsoft:
Is there a reason to keep getProperty in class HystrixCollapserProperties with '@SuppressWarnings("unused")' annotation?

mattrjacobs:
No good reason.  Removed in #1151

fcalderonnearsoft:
@mattrjacobs I found more https://github.com/Netflix/Hystrix/pull/1154

mattrjacobs:
Closed via #1154

PR #1154, opened 2016-03-17
Removed methods and variables with unused annotation

fcalderonnearsoft:
I found more unused methods and variables related with  https://github.com/Netflix/Hystrix/issues/1150

### DIFF
--- a/hystrix-contrib/hystrix-metrics-event-stream/src/main/java/com/netflix/hystrix/contrib/metrics/eventstream/HystrixMetricsPoller.java
+++ b/hystrix-contrib/hystrix-metrics-event-stream/src/main/java/com/netflix/hystrix/contrib/metrics/eventstream/HystrixMetricsPoller.java
@@ -134,6 +134,19 @@ public class HystrixMetricsPoller {
     /**
      * Used to protect against leaking ExecutorServices and threads if this class is abandoned for GC without shutting down.
      */
+    @SuppressWarnings("unused")
+    private final Object finalizerGuardian = new Object() {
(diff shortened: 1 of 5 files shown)


fcalderonnearsoft:
Hi @mattrjacobs when you have time could you check this, please?

hystrix-contrib/hystrix-metrics-event-stream/src/main/java/com/netflix/hystrix/contrib/metrics/eventstream/HystrixMetricsPoller.java
```diff
@@ -134,19 +134,6 @@ public boolean isRunning() {
     /**
      * Used to protect against leaking ExecutorServices and threads if this class is abandoned for GC without shutting down.
      */
-    @SuppressWarnings("unused")
```
mattrjacobs: This one does get called, just not within the Hystrix library.  See http://stackoverflow.com/questions/6872857/how-does-a-finalizer-guardian-work-in-java for a full explanation of how this works.  This one should stay.

mattrjacobs:
@fcalderonnearsoft Sorry for the delay - was out on vacation for a few days.  I left a note on the finalizer guardian, but the others look fine to remove.  If you can re-add the finalizer guardian, I will merge the PR.  Thanks for the contribution!

fcalderonnearsoft:
@mattrjacobs  No problem, I already re-add finalizerguardian in a new commit. 

Cheers.

PR #1156, opened 2016-03-18
Hystrix Metrics Stream with Vertx

kennedyoliveira:
Hello there!

This PR is a implementation of a metrics stream with Vertx, is fast, efficient, asynchronous, use lower resources, and is very easily to use and to deploy. Don't need a servlet container too.

I hope it can help!


kennedyoliveira:
Didn't saw the project was using jdk7 for building and source/target java6, vertx just works on java 8 :(
Nothing to do?

mattrjacobs:
@kennedyoliveira This sounds great, but Java8 is not planned for all of Hystrix in any of the 1.x releases.  The current situation is:

* All Hystrix modules, including `hystrix-core` are JDK6.
* Travis doesn't support JDK6 for building (AFAIK), so it uses JDK7.

I checked and it's possible to change the source/target JDKs for specific Gradle subprojects.  Just add

```
sourceCompatibility = 1.8
targetCompatibility = 1.8
```

to the `build.gradle` in the submodule you want to be Java 8.

kennedyoliveira:
hey @mattrjacobs, thx for the feedback, i know i can change the source and target, actually my commit this options is already set, the problem is the travis build is using jdk 7 to build, can we change to use jdk8 to build? And your cloudbees are using jdk 7 too, so it gives the error.
I think there is no problem since all the projects and subprojects are set to jdk 6 so build it from 8 makes no difference, and will allow my module to be built too.
What you say?

agentgt:
@kennedyoliveira the easiest solution would be to just create a separate project and then just depend on what ever version of Hystrix is required. 

To be honest I think there are almost too many modules in Hystrix contrib already (IMO the javana one being a close example). Vert.x is fairly opinionated and not a common component (e.g. it would be like having a Play framework plugin for Hystrix in contrib).

Also If I recall Vert.x has its own plugin repository which might be a better home.

kennedyoliveira:
@agentgt, ok, i'll create another project for it.

mattrjacobs:
I agree with @agentgt that the division between projects that consume Hystrix are are in a contrib module or are in an external project are fairly arbitrary at this point.  I'm working on finding a solution to this and should have more to publicly share this week.  

In this specific case, given that no internal Hystrix contributors are Vert.x experts, means that supporting this project would likely be something I have to find someone to do. 

The Vert.x plugin sounds like a very good idea, IMO.

kennedyoliveira:
@mattrjacobs, i created a separate project, if you have any interest [hystrix-vertx-metrics-stream](https://github.com/kennedyoliveira/hystrix-vertx-metrics-stream)

mattrjacobs:
@kennedyoliveira Nice work!

kennedyoliveira:
@mattrjacobs, Thank you!

Issue #1155, opened 2016-03-17
Hystrix exceptions not passed into fallback.


ckstettler:
I have implemented a fallback that accepts Throwable.

What I see is the following:

If I get into my @HystrixCommand annotated function and throw an exception, then I do in fact see the correct throwable in the fallback function.

If the exception is raised as part of the @HystrixCommand processing within javanica such as timeout or queue threshold is reached etc, then the throwable in the fallback is not populated.

Is there anyway for me to see these exceptions?  I am using 1.5.1.

mattrjacobs:
@ckstettler Thanks for the report.  Can someone more familiar with `hystrix-javanica` look into this bug? 
/cc @dmgcodevil

dmgcodevil:
Hi, it's a duplicate https://github.com/Netflix/Hystrix/issues/974
Problem was getFailedExecutionException() doesn't return exceptions in case of timeouts and etc but nulls. to get exception javanica uses getFailedExecutionException().
I remember Matt fixed this issue in core. 
@mattrjacobs  should I change something in javanica to get such exceptions and propogate them ?

mattrjacobs:
@dmgcodevil IIRC, it's the change you mentioned here: https://github.com/Netflix/Hystrix/issues/1013#issuecomment-163368438

dmgcodevil:
Oh, right, will change today then

fayndee:
Hi @dmgcodevil, is there any progress on this fix?

dmgcodevil:
Hi @fayndee, yes, Spencer has fixed the issue. I guess we can close this issue

mattrjacobs:
Released in 1.4.26/1.5.2

ckstettler:
Awesome!! Thank you.

PR #1167, opened 2016-04-01
Use getExecutionException to handle timeout, short-circuit, etc... ex…

spencergibb:
…ceptions.

fixes gh-1155

/cc @mattrjacobs @dmgcodevil 

Is there anything specific I should do to test this?


mattrjacobs:
Thanks @spencergibb .  I'll also backport this to 1.4.x so the fix is available there

PR #1181, opened 2016-04-15
Deprecate getter and setter for unused collapsingEnabled property in Collapser Setter.

nluchs:
I might be mistaken, but it looks like the "collapsingEnabled" property in HystrixCollapserProperties.Setter doesn't ever get used, which is kind of confusing, since I had assumed it would actually enable/disable collapsing. So I figure a straightforward way to avoid people using it in the future is Deprecation in current versions and removing it 2.x or whatever. Thoughts? I can submit another pull request to remove it from the 2.x branch if that's a good idea.

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/HystrixCollapserProperties.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/HystrixCollapserProperties.java
@@ -213,7 +213,7 @@ public abstract class HystrixCollapserProperties {
      * @NotThreadSafe
      */
     public static class Setter {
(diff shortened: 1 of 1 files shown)


mattrjacobs:
Yeah, nice catch, @nluchs .  Would be happy to include a deletion in the 2.x branch as well.  Thanks!

mattrjacobs:
@nluchs Headsup that I did the deletion already in 2.0.x (#1187).  Wanted to make sure I didn't forget to.  Thanks again!

nluchs:
Sounds good, thanks Matt!

PR #1184, opened 2016-04-20
migrating all hystrix-javanica tests to hystrix-junit

caarlos0:
using the rule created on #1147 on all hystrix-javanica tests.

Also improved the rule a bit.

cc/ @mattrjacobs


mattrjacobs:
Thanks @caarlos0

PR #1192, opened 2016-04-21
Added AbstractCollapser

rdegnan:
Consolidate a lot of the copy-pasted code in HystrixCollapser and HystrixObservableCollapser into one AbstractCollapser class and allow HystrixCollapser to be used with HystrixObservableCommands.

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/AbstractCollapser.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/AbstractCollapser.java
@@ -1,0 +1,427 @@
+/**
+ * Copyright 2012 Netflix, Inc.
+ * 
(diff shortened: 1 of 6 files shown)


mattrjacobs:
Just ran migration-manager and found:

```
Found 2 binary incompatibilities
================================
 * abstract method
   mapResponseToRequests(rx.Observable,java.util.Collection)rx.Observable in
   interface com.netflix.hystrix.collapser.HystrixCollapserBridge does not have
   a correspondent in current version
 * abstract method
   handleResponse(rx.Observable,java.util.Collection)rx.Observable in interface
   com.netflix.hystrix.collapser.HystrixCollapserBridge is present only in
   current version
```

Are you able to fix that?

PR #1193, opened 2016-04-24
more tests using HystrixRequestContext rule

caarlos0:
Replaced the manual `HystrixRequestContext` inits and shutdowns with the junit rule.

My IDE also auto-fixed some trailing whitespaces, that's why the pull seems a little big. A `diff -w` should do the trick.

refs https://github.com/Netflix/Hystrix/pull/1184 and https://github.com/Netflix/Hystrix/pull/1147

cc/ @mattrjacobs

### DIFF
--- a/hystrix-core/build.gradle
+++ b/hystrix-core/build.gradle
@@ -7,6 +7,7 @@ dependencies {
     compile 'org.slf4j:slf4j-api:1.7.0'
 	compile 'org.hdrhistogram:HdrHistogram:2.1.7'
     testCompile 'junit:junit-dep:4.10'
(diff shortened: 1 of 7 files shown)


mattrjacobs:
Thanks @caarlos0!  This looks like it should introduce a circular dependency where `hystrix-junit:compile` depends on `hystrix-core:compile`, and `hystrix-core:test` depends on `hystrix-junit:compile`.  From running tests locally and in CI, it looks like Gradle is able to sidestep the circularity.

caarlos0:
@mattrjacobs yeap, gradle seems to be smart enough to handle that...

PR #1194, opened 2016-04-24
improved tests readbility a bit

caarlos0:
all tests were doing stuff like `try { something(); } catch (Exception e) { fail(); }`, I removed those try-catch blocks were I thought was safe to.

I also removed some commented out code..


mattrjacobs:
Seems like a good change to me as well.  Thanks, @caarlos0 !

caarlos0:
thanks @mattrjacobs :)

Issue #1199, opened 2016-05-05
HystrixEventType.UNSUBSCRIBED


mattrjacobs:
Currently, if an Observable produced by Hystrix gets subscribed to, and then unsubscribed before any emissions, then a lot of state is uninitialized.

Specifically, the list of events is empty and the latency is -1.  I believe we should have a separate event type when unsubscription happens.  IMO, we should omit it when there is a terminal state, but include it otherwise.

mdjnewman:
I see this has been resolved for a while - does the documentation at https://github.com/Netflix/Hystrix/wiki/Metrics-and-Monitoring need updating to include `CANCELLED`? :)

Issue #220, opened 2014-03-13
Hystrix Command should be cancellable


panterose:
At the moment, if you call .queue() on a command and get a future, you cannot call the cancel() method and abort its execution. This is true for the current 1.3.XX version (it seems to work if you don't use the cache for some reason) but also in the upcoming 1.4 version.

I can supply a reduced code sample if needed, but it's very easy to reproduced. It's an issue for me as we have some long running command.

We also have the case where we encapsulate child command inside main command, ideally the cancel() should be cascading.

mattrjacobs:
Talked to @benjchristensen about this.  Since the semantics are slightly different than ```Subject.refCount()```, it will take some time/care to do this properly.  As a result targeting this for 1.4 series, but not gating RC7 on it.

mattrjacobs:
This is fixed in #1204.  Will be released in 1.5.3

PR #1204, opened 2016-05-11
Support unsubscription of HystrixCommands

mattrjacobs:
This is a large reorganization of Hystrix internals to support unsubscription.

The changes are:
* Move cleanup actions to happen on the first of terminate/unsubscribe
* Wire up Hystrix state machine as a sequence of operators, rather than using explicit `Observable.create`
* Be explicit about which state is shared across commands in a request (HystrixCommand.executionResult), and that which is owned by a single command.  For example, a command coming from cache, or being cancelled, is only a property of that command, while a SUCCESS is a property of all.

I've added a number of tests to check unsubscription.  Outstanding work:

* Let explicit unsubscription trigger a thread interruption.  At the moment, only timeouts trigger thread interruptions, while explicit unsubscription triggers a cancellation of the future without a corresponding interrupt.
* Add CANCELLED metrics
* At least one edge case exists for metrics hat isn't handled well at the moment: Command A starts execution, then Command B gets invoked with the same arguments and is supplied by the Observable in Command A.  Then, Command A is unsubscribed.  The underlying work still goes on, but A's latency stops when it is unsubscribed.  Since Command B comes from cache, its latency is 0.  So nothing tracks the remainder of the underlying work from when A gets cancelled to its completion.


mattrjacobs:
This addresses #1199 and #220

stevegury:
👍 
No big deal but a couple of those new Atomic variables could be replaced by `volatile` variables.

mattrjacobs:
Thanks @stevegury.  Just modified the state in `AbstractCommand` to be volatile whenever possible

stevegury:
👍

mmanciop:
I think the implementation of com.netflix.hystrix.HystrixCommand.queue() is still not compliant with Future.cancel(boolean) as it does not seem to interrupt the execution thread.

mattrjacobs:
I just created a new issue to track making this work properly: #1268

PR #1210, opened 2016-05-21
no need to fetch the method's parameter annotations, if there aren't any

mebigfatguy:


mattrjacobs:
Thanks for the contribution, @mebigfatguy

PR #1211, opened 2016-05-23
ReactiveSocket metrics stream

robertroeser:
Metrics stream implementation in reactive sockets. Takes a 32-bit integer that represents the stream type. Data is streamed as CBOR.

### DIFF
--- a/hystrix-contrib/hystrix-reactivesocket-event-stream/src/main/java/com/netflix/hystrix/contrib/reactivesocket/BasePayloadSupplier.java
+++ b/hystrix-contrib/hystrix-reactivesocket-event-stream/src/main/java/com/netflix/hystrix/contrib/reactivesocket/BasePayloadSupplier.java
@@ -1,0 +1,19 @@
+package com.netflix.hystrix.contrib.reactivesocket;
+
+import com.fasterxml.jackson.dataformat.cbor.CBORFactory;
+import io.reactivesocket.Payload;
+import rx.Observable;
+import rx.subjects.BehaviorSubject;
+
+import java.util.function.Supplier;
+
+public abstract class BasePayloadSupplier implements Supplier<Observable<Payload>> {
+    protected final CBORFactory jsonFactory;
+
+    protected final BehaviorSubject<Payload> subject;
+
+    protected BasePayloadSupplier() {
+        this.jsonFactory = new CBORFactory();
+        this.subject = BehaviorSubject.create();
+    }
+}
--- a/hystrix-contrib/hystrix-reactivesocket-event-stream/src/main/java/com/netflix/hystrix/contrib/reactivesocket/StreamingSupplier.java
+++ b/hystrix-contrib/hystrix-reactivesocket-event-stream/src/main/java/com/netflix/hystrix/contrib/reactivesocket/StreamingSupplier.java
@@ -1,7 +1,77 @@
 package com.netflix.hystrix.contrib.reactivesocket;
 
-/**
- * Created by rroeser on 5/18/16.
- */
-public class StreamingSupplier {
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonGenerator;
+import io.reactivesocket.Frame;
+import io.reactivesocket.Payload;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import rx.Observable;
+import rx.functions.Func0;
+import rx.schedulers.Schedulers;
+import rx.subjects.BehaviorSubject;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
+
+public abstract class StreamingSupplier<T> implements Supplier<Observable<Payload>> {
+
+    protected Logger logger = LoggerFactory.getLogger(StreamingSupplier.class);
+
+    protected BehaviorSubject<Payload> subject;
+
+    protected final JsonFactory jsonFactory;
+
+    protected StreamingSupplier() {
+        subject = BehaviorSubject.create();
+        jsonFactory = new JsonFactory();
+
+        Observable
+            .interval(500, TimeUnit.MILLISECONDS, Schedulers.computation())
+            .doOnNext(i ->
+                getStream()
+                    .filter(this::filter)
+                    .map(this::getPayloadData)
+                    .forEach(b -> {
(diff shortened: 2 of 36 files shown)


hystrix-contrib/hystrix-reactivesocket-event-stream/src/main/java/com/netflix/hystrix/contrib/reactivesocket/metrics/HystrixCollasperMetricsStream.java
```diff
@@ -0,0 +1,106 @@
...
+import com.netflix.hystrix.HystrixCollapserMetrics;
+import com.netflix.hystrix.HystrixEventType;
+import com.netflix.hystrix.contrib.reactivesocket.StreamingSupplier;
+import org.agrona.LangUtil;
+import rx.functions.Func0;
+
+import java.io.ByteArrayOutputStream;
+import java.util.stream.Stream;
+
+public class HystrixCollasperMetricsStream extends StreamingSupplier<HystrixCollapserMetrics> {
```
mattrjacobs: Typo: Collasper -> Collapser

hystrix-contrib/hystrix-reactivesocket-event-stream/src/main/java/com/netflix/hystrix/contrib/reactivesocket/BasePayloadSupplier.java
```diff
@@ -0,0 +1,19 @@
...
+
+import com.fasterxml.jackson.dataformat.cbor.CBORFactory;
+import io.reactivesocket.Payload;
+import rx.Observable;
+import rx.subjects.BehaviorSubject;
+
+import java.util.function.Supplier;
+
+public abstract class BasePayloadSupplier implements Supplier<Observable<Payload>> {
+    protected final CBORFactory jsonFactory;
```
mattrjacobs: should this be named cborFactory?

hystrix-contrib/hystrix-reactivesocket-event-stream/src/main/java/com/netflix/hystrix/contrib/reactivesocket/StreamingSupplier.java
```diff
@@ -0,0 +1,70 @@
...
+                            @Override
+                            public ByteBuffer getMetadata() {
+                                return Frame.NULL_BYTEBUFFER;
+                            }
+                        };
+
+                        subject.onNext(p);
+                    })
+            )
+            .retry()
```
mattrjacobs: What role does `retry` play here?

hystrix-contrib/hystrix-reactivesocket-event-stream/src/test/java/com/netflix/hystrix/contrib/reactivesocket/EventStreamEnumTest.java
```diff
@@ -0,0 +1,8 @@
+package com.netflix.hystrix.contrib.reactivesocket;
+
+
+public class EventStreamEnumTest {
```
mattrjacobs: Delete this class?

hystrix-contrib/hystrix-reactivesocket-event-stream/src/test/java/com/netflix/hystrix/contrib/reactivesocket/metrics/HystrixCollasperMetricsStreamTest.java
```diff
@@ -0,0 +1,49 @@
+package com.netflix.hystrix.contrib.reactivesocket.metrics;
+
+import com.netflix.hystrix.HystrixCommand;
+import com.netflix.hystrix.HystrixCommandGroupKey;
+import org.junit.Test;
+
+import java.nio.ByteBuffer;
+import java.util.concurrent.CountDownLatch;
+
+public class HystrixCollasperMetricsStreamTest {
```
mattrjacobs: Typo: Collasper -> Collapser

hystrix-contrib/hystrix-reactivesocket-event-stream/src/main/java/com/netflix/hystrix/contrib/reactivesocket/StreamingSupplier.java
robertroeser: Don't want this to unsubscribe if there is an error

hystrix-contrib/hystrix-reactivesocket-event-stream/src/main/java/com/netflix/hystrix/contrib/reactivesocket/BasePayloadSupplier.java
robertroeser: maybe - CBORFactory inherits from jsonFactory though and you can replace it with the JsonFactory if you want.

mattrjacobs:
A couple more questions:

1) How should this actually get exposed over the network?  As an example, if I wanted to use this for `hystrix-examples-webapp`, I could set up a `EventStreamRequestHandler`.  From that, I can get a `Publisher<Payload>` by passing in the appropriate enum.  But I'm not familiar enough with ReactiveSocket to know how to expose that, and then set up a client on the other side.

2) The code to convert data types into JSON/CBOR is very similar to `hystrix-metric-event-stream`.  Would it make sense to directly expose a map-like object from `hystrix-core`, on the assumption that any serializer wants that?  If that were exposed, then any given serializer wouldn't care anything about the content of the data, just the field-types.

robertroeser:
_1) How should this actually get exposed over the network? As an example, if I wanted to use this for hystrix-examples-webapp, I could set up a EventStreamRequestHandler. From that, I can get a Publisher<Payload> by passing in the appropriate enum. But I'm not familiar enough with ReactiveSocket to know how to expose that, and then set up a client on the other side._

EventStreamRequestHandler is roughly equivalent to a HttpServlet. When you create an instance of a ReactiveSocket you would give it an instance of this class.

_2) The code to convert data types into JSON/CBOR is very similar to hystrix-metric-event-stream. Would it make sense to directly expose a map-like object from hystrix-core, on the assumption that any serializer wants that? If that were exposed, then any given serializer wouldn't care anything about the content of the data, just the field-types._
Sure - add a method on the Metrics objets that gets the metrics as a map?

mattrjacobs:
Merging this, and I'll work on building a client and the data type work outlined in my second point above.

PR #1213, opened 2016-05-24
use parameterized logging

mebigfatguy:
avoid String building if the log level doesn't cause the line to be logged.


mattrjacobs:
Thanks, @mebigfatguy !

PR #1219, opened 2016-05-26
avoid map lookups, in loops

mebigfatguy:


mattrjacobs:
Thanks, @mebigfatguy !

PR #1221, opened 2016-05-26
improve usages of "its" and "it's"

dysmento:
Hi. This is a small thing, but I noticed this grammatical error in the logs and thought I could fix it in a jiffy.


mattrjacobs:
Thanks @dysmento !

PR #1229, opened 2016-05-31
remove dead Setter class

mebigfatguy:


mattrjacobs:
Thanks @mebigfatguy !

PR #1231, opened 2016-06-03
make inner classes static where possible, and remove outer class refe…

mebigfatguy:
…rence


mattrjacobs:
Thanks @mebigfatguy !

PR #1232, opened 2016-06-03
compare enums with ==, avoid chances of NPE

mebigfatguy:


mattrjacobs:
Good idea, thanks @mebigfatguy !

PR #1236, opened 2016-06-10
ReactiveSocket metrics client/server

mattrjacobs:
* Client/Server in hystrix-examples-reactivesocket
* Improvements to HystrixCommandAsyncDemo
* Bugfixes/polishing to hystrix-reactivesocket-metric-stream.  Sources should support arbitrary sharing between 0-n subscribers that may come and go.  When subscriber count drops to 0, source does no work


hystrix-contrib/hystrix-reactivesocket-event-stream/src/main/java/com/netflix/hystrix/contrib/reactivesocket/serialize/SerialHystrixConfiguration.java
```diff
@@ -98,6 +83,108 @@ public static HystrixConfigStream getInstance() {
...
+                                metricsConfig.path("percentileBucketSizeInMilliseconds").asInt(),
+                                metricsConfig.path("percentileEnabled").asBoolean(),
+                                metricsConfig.path("counterBucketCount").asInt(),
+                                metricsConfig.path("counterBucketSizeInMilliseconds").asInt()
+                        )
+                );
+                collapserConfigMap.put(collapserKey, collapserConfig);
+            }
+        } catch (IOException ioe) {
+            System.out.println("IO Exception : " + ioe);
```
nickmahilani: can log with logger here

mattrjacobs:
Good catch @nickmahilani - just found 2 others as well

hystrix-contrib/hystrix-reactivesocket-event-stream/src/main/java/com/netflix/hystrix/contrib/reactivesocket/EventStream.java
```diff
@@ -0,0 +1,92 @@
...
+                        .map(SerialHystrixMetric::toPayload);
+                break;
+            case UTILIZATION_STREAM:
+                source = new HystrixUtilizationStream(UTILIZATION_DATA_INTERVAL_IN_MS)
+                        .observe()
+                        .map(SerialHystrixUtilization::toBytes)
+                        .map(SerialHystrixMetric::toPayload);
+                break;
+            case GENERAL_DASHBOARD_STREAM:
+                source = HystrixDashboardStream.getInstance()
```
nickmahilani: For my understanding, why is the dashboard stream delay not configurable?

mattrjacobs:
Right now, the utilization and configuration streams always accept a stream delay and are non-singletons.  Looking back, this is probably a mistake as it makes it harder for multiple subscribers to share.

I think there are 2 more tasks (coming next week):
1. Make util/config streams singletons
2. Hook the singleton streams up to Archaius so that there's a way to do something other than what we hard-code.  I think that should still require a restart though.

nickmahilani:
just couple minor comments, PR looks good to me 👍

hystrix-contrib/hystrix-reactivesocket-event-stream/src/main/java/com/netflix/hystrix/contrib/reactivesocket/EventStream.java
nickmahilani: Singletons for util/config stream to allow subscribers to share w/ archaius sounds like a good approach.

PR #1244, opened 2016-06-14
fix handling of invalid weavingMode property

mebigfatguy:
Enum.valueOf throws an exception, and does not return null, so handle invalid value that way.


mattrjacobs:
Nice catch - thanks @mebigfatguy

PR #1247, opened 2016-06-16
Set the thread pool size only when the configuration is changed.

yanglifan:
This contribution will make Hystrix run faster in JDK 6.  Since when create a `HystrixCommand`, `HystrixThreadPoolDefault.touchConfig` will be executed, then execute`ThreadPoolExecutor.setCorePoolSize` and `ThreadPoolExecutor.setMaximumPoolSize`.  These two methods will execute `mainLock.lock()` all the time.  That is negative to the performance.

The improve is to check whether pool size will be changed or not before set this value.

This contribution will not affect JDK 7 and 8.


mattrjacobs:
Thanks, @yanglifan!  I agree that the CI failure is unrelated.  I'll merge in your PR and then get that unit test addressed

Issue #1251, opened 2016-06-18
HystrixDashboard - doesn't seem to display the right TPS


jmirc:
I am using the version 1.1.2.RELEASE of turbine stream and I think the dashboard has an issue. Let me know, if I am wrong or not.

Here is an example of the data that I got from the turbine stream

```
[...]

PR #1253, opened 2016-06-21
Fix issue #1251

jmirc:


mattrjacobs:
Thanks, @jmirc

raveeshrana:
@jmirc @mattrjacobs 
Where can I find more details on why this was changed? 

I am getting accumulated property Value_metricsRollingStatisticalWindowInMilliseconds values and it must be averaged out perHost to get the right result.

raveeshrana:
@mattrjacobs @jmirc 
I am getting aggregated value from propertyValue_metricsRollingStatisticalWindowInMilliseconds.

What could I be doing different .. Also I see similar code in this file as well .. Do they look out of sync ? https://github.com/yammer/breakerbox/blob/ac6960bf28509fb04198409bdae9b97c0e29e260/breakerbox-dashboard/src/main/resources/assets/components/hystrixThreadPool/hystrixThreadPool.js#L97

Issue #1177, opened 2016-04-13
Javanica fallback not found when contained in super class


ctoestreich:
We are using the Annotated methods on classes where a service extends another service

```groovy 
class WishlistService extends CartService { ... }
```

And then inside CartService there is a some methods like

``` groovy 
 @HystrixCommand(commandProperties = [
            @HystrixProperty(name = "execution.isolation.thread.timeoutInMilliseconds", value = "5000")
    ], threadPoolProperties = [
            @HystrixProperty(name = "coreSize", value = "20"),
            @HystrixProperty(name = "maxQueueSize", value = "150"),
            @HystrixProperty(name = "queueSizeRejectionThreshold", value = "150")
    ], threadPoolKey = 'CartService', fallbackMethod = 'transferItemsFallback')
    CartResponseDTO transferItems(CartKey fromKey, CartKey toKey, List itemIds){
     ...
}

CartResponseDTO transferItemsFallback(CartKey fromKey, CartKey toKey, List items, Throwable e) { ... }
```

When using the service `wishListService` in a class the following exception is thrown

``` console
com.netflix.hystrix.contrib.javanica.exception.FallbackDefinitionException: fallback method wasn't found: transferItemsFallback([class com.bsb.cart.domain.CartKey, class com.bsb.cart.domain.CartKey, interface java.util.List])
	at com.netflix.hystrix.contrib.javanica.utils.MethodProvider.getFallbackMethod(MethodProvider.java:82) ~[hystrix-javanica-1.5.2.jar:1.5.2]
	at com.netflix.hystrix.contrib.javanica.utils.MethodProvider.getFallbackMethod(MethodProvider.java:55) ~[hystrix-javanica-1.5.2.jar:1.5.2]
	at com.netflix.hystrix.contrib.javanica.aop.aspectj.HystrixCommandAspect$MetaHolderFactory.metaHolderBuilder(HystrixCommandAspect.java:120) ~[hystrix-javanica-1.5.2.jar:1.5.2]
	at com.netflix.hystrix.contrib.javanica.aop.aspectj.HystrixCommandAspect$CommandMetaHolderFactory.create(HystrixCommandAspect.java:208) ~[hystrix-javanica-1.5.2.jar:1.5.2]
	at com.netflix.hystrix.contrib.javanica.aop.aspectj.HystrixCommandAspect$MetaHolderFactory.create(HystrixCommandAspect.java:106) ~[hystrix-javanica-1.5.2.jar:1.5.2]
[...]

PR #1258, opened 2016-06-26
iss1177: change getMethod to recursively search in parent types

dmgcodevil:
link https://github.com/Netflix/Hystrix/issues/1177


mattrjacobs:
Unrelated test failure on one of my recent changes.  Merging and will investigate that separately

Issue #1228, opened 2016-05-31
Unable to set threadPoolProperties using Javanica on HystrixObservableCommand


yesdave:
If I annotate my Observable returning command with `threadPoolProperties`:

```java
@HystrixCommand(
    commandProperties = {
        @HystrixProperty(name = "execution.isolation.strategy", value = "THREAD")
    },
    threadPoolProperties = {
        @HystrixProperty(name = "coreSize", value = "20")
    }
)
public Observable<User> getUser(Long id) {
      ...
}
```

It appears to be ignored:
> java.util.concurrent.FutureTask@7b9f4fb4 rejected from java.util.concurrent.ThreadPoolExecutor@69f1dd4f[Running, pool size = 10, active threads = 10, queued tasks = 0, completed tasks = 0]

This looks to be because `GenericSetterBuilder.buildObservableCommandSetter()` only sets commandProperties and ignores threadPoolProperties, unlike the `build()` method, which in turn might be because (core) `HystrixObservableCommand.Setter` class only deals with commandProperties.

Have threadPoolProperties been intentionally ignored on Observables, has this not been implemented yet, or am I missing something else entirely?

mattrjacobs:
Allowing `HystrixObservableCommand`s to use thread-isolation was a mistake, and will be removed in a future version of Hystrix.  See #1200 for more details.

The 2 recommended ways to use Hystrix are:
1) If you are wrapping blocking I/O, model this as a `HystrixCommand` (not `HystrixObservableCommand`).  Use thread-isolation for resilience benefits.  You also gain the ability to perform async execution via `queue()`/`observe()`/`toObservable()`.
2) If you are wrapping nonblocking I/O, model this as a `HystrixObservableCommand`.  Use semaphore-isolation.

swapnil1988:
I am evaluating hystrix for my new project. Correct  me if I am wrong but it seems I have a requirement where in I would require thread isolation for hystericObservableCommand. 
We would be using vertx framework along with hystrix. Vertx is non blocking async framework but also support blocking execution. So we would be using mix of both blocking and non blocking code. 
As part of wrapping a service call in hystrix we would have some blocking code as well as some non blocking in same service. Example would be some blocking code processing using legacy java library then send a message using vertx event bus and pass handler for getting reply that is async/ non blocking and when reply is received service execution is complete. 
Do let me know if what ever I have thought of make sence ?

Issue #1200, opened 2016-05-05
Eliminate thread-isolated HystrixObservableCommands


mattrjacobs:
After writing a few HystrixObservableCommands in real systems, I've come to the conclusion that we should not support thread-isolation for HystrixObservableCommands.  Here's why:

In a `HystrixCommand`, Hystrix is responsible for turning a function returning `T` into `Observable<T>`.  Hystrix will call `subscribeOn` on this Observable to move it to a Hystrix thread if the command is thread-isolated.

In a `HystrixObservableCommand`, Hystrix is given a function that already returns an `Observable<T>`.  Hystrix doesn't know anything about where the source observable runs.  It might already have a `subscribeOn` in its observable chain, in which case any further `subscribeOn`s will be no-ops.  This implies that Hystrix can not always perform thread-isolation.

Instead, we should trust that the user has already decided where the source observable should run, and apply all of the Hystrix logic to that thread.  This is semaphore-isolation, and it's, IMO, the only reasonable choice for `HystrixObservableCommand`s.

This will change some semantics, so I want to make this 1.6.0.

yanglifan:
So in Hystrix 1.6, the execution in a `HystrixObservableCommand` will only be semaphore isolated?

mattrjacobs:
I haven't made this change yet, so I don't know in which version this will land.  I'm planning on spending some time rounding up all the current/planned deprecations and mapping the deprecation/deletions to releases.  Will ping this thread when I have that.

swapnil1988:
I am evaluating hystrix for my new project. Correct me if I am wrong but it seems I have a requirement where in I would require thread isolation for hystericObservableCommand. 
We would be using vertx framework along with hystrix. Vertx is non blocking async framework but also support blocking execution. So we would be using mix of both blocking and non blocking code. 
As part of wrapping a service call in hystrix we would have some blocking code as well as some non blocking in same service. Example would be some blocking code processing using legacy java library then send a message using vertx event bus and pass handler for getting reply that is async/ non blocking and when reply is received service execution is complete. 
In such case won't it be good to use hystrixObservableCommand with thread isolation as my  service would have both blocking and non blocking code.
Do let me know if what ever I have thought of make sence ?

mattrjacobs:
If you have blocking I/O, use a thread-isolated HystrixCommand.
If you have nonblocking I/O, use a semaphore-isolated HystrixObservableCommand.  

Any concerns with that approach?

[...]

PR #1264, opened 2016-06-30
Allow to configure threadPoolKey through HystrixObservableCommand.Setter

heiko-braun:
Provides the equivalent of HystrixCommand.Setter


mattrjacobs:
This is not desirable.  See previous discussion here: https://github.com/Netflix/Hystrix/issues/805#issuecomment-109371037
https://github.com/Netflix/Hystrix/issues/1228#issuecomment-222823739

In #1200, I plan to remove thread-isolated HystrixObservableCommands.

heiko-braun:
Thanks, makes complete sense to me.

PR #1265, opened 2016-06-30
Interrupt execution thread on HystrixCommand#queue()#cancel(true)

mmanciop:
The implementation of java.util.concurrent.Future returned by com.netflix.hystrix.HystrixCommand#queue()#cancel(boolean mayInterruptIfRunning) ignores the "mayInterruptIfRunning" flag and never calls Thread#interrupt() on Hystrix Execution Thread running the command's run() method. This complicates the handling of situations where commands have to be interrupted. I am not sure it would make sense to enable the interruption of the getFallback() method, as well-behaved fallbacks probably should not do stuff that may require interruption.

The interruption of the execution thread IMO makes sense only through the invocation of the Future#cancel(boolean) method. The command should not be interruptible if executed via com.netflix.hystrix.HystrixCommand#execute() (blocking invocation forfeits interruption) or com.netflix.hystrix.AbstractCommand#observe() (interruption does not seem to me to be a concept you'd want to mix in a pure reactive consumption of a command; for that, you could check for unsubscription). As the current behavior is never to interrupt the running thread, we should probably hide the new behavior behind a command property with default false to preserve the previous behavior.

[Please notice that the change linked in this pull request is alpha-quality at best and is meant to collect feedback on (1) if it makes sense at all to work on this and (2) if there are better places where the lifecycle can be extended to obtain the intended result:-) ]

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/HystrixCommand.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/HystrixCommand.java
@@ -17,10 +17,13 @@ package com.netflix.hystrix;
 
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 
 import rx.Observable;
-import rx.Observable.OnSubscribe;
-import rx.Subscriber;
+import rx.functions.Action0;
 
 import com.netflix.hystrix.exception.HystrixBadRequestException;
 import com.netflix.hystrix.exception.HystrixRuntimeException;
@@ -257,7 +260,10 @@ public abstract class HystrixCommand<R> extends AbstractCommand<R> implements Hy
 
     }
 
-    /**
+	private final AtomicReference<Thread> executionThread = new AtomicReference<Thread>();
+	private final AtomicBoolean interruptOnFutureCancel = new AtomicBoolean(false);
+
+	/**
      * Implement this method with code to be executed when {@link #execute()} or {@link #queue()} are invoked.
      * 
      * @return R response type
@@ -294,6 +300,12 @@ public abstract class HystrixCommand<R> extends AbstractCommand<R> implements Hy
                 } catch (Throwable ex) {
                     return Observable.error(ex);
                 }
+            }
+        }).doOnSubscribe(new Action0() {
+            @Override
+            public void call() {
+                // Save thread on which we get subscribed so that we can interrupt it later if needed
+                executionThread.set(Thread.currentThread());
             }
         });
     }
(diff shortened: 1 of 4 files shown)


mmanciop:
On second thought, if someone used Future.cancel(true) on the Future returned by HystrixCommand.queue(), they did want the interruption to occur, so hiding this behavior behind a configuration property adds unnecessary complexity.

mattrjacobs:
@mmanciop Thanks for the PR.  Will review it shortly.  Addresses #1268

mattrjacobs:
This looks really nice.  Your choice to only allow the interruption on explicit HystrixCommand.queue() is a very good one, IMO.  A couple of comments inline

hystrix-core/src/test/java/com/netflix/hystrix/HystrixCommandTest.java
```diff
@@ -2766,6 +2766,36 @@ public void testDoNotInterruptToObservableOnTimeoutIfPropertySaysNotTo() throws
     }
 
     @Test
```
mattrjacobs: Can you also add a test when cancel(false) is called?  In this case, I would expect the Hystrix thread to keep sleeping, and the latch timeout to get hit in the unit test

hystrix-core/src/main/java/com/netflix/hystrix/HystrixCommand.java
```diff
@@ -370,7 +399,42 @@ public R execute() {
...
+        final Future<R> f = new Future<R>() {
+        	final Future<R> base = o.toBlocking().toFuture();
+
+        	@Override
+			public boolean cancel(boolean mayInterruptIfRunning) {
+        		if (base.isCancelled()) {
+        			return false;
+        		}
+
+        		mayInterruptOnCancel.set(mayInterruptIfRunning);
```
mattrjacobs: I think the only valid transition here is false -> true.  If a user does something really weird and gets 2 futures out (f1 and f2), then calls f1.cancel(true); f2.cancel(false), it's unclear about what value would be used by the time `mayInterruptOnCancel` is checked.  I would want the value set to true in that case.

So could you do something like:

```
if (mayInterruptIfRunning) {
    mayInterruptOnCancel.set(true);
}
```

mmanciop:
Thanks for reviewing! I currently have no access to any decent network connectivity. I'll address your comments around the July 25 or so. Cheers.

mattrjacobs:
No problem.  I'm taking vacation soon, so I can review in August when I get back

hystrix-core/src/test/java/com/netflix/hystrix/HystrixCommandTest.java
mmanciop: There :-)

mmanciop:
Quite frustratingly, the build and tests succeed consistently in local. From Travis I am not even sure if the new tests for this pull request are the ones causing the issue with the build... Any pointers?

mattrjacobs:
I've spent some time getting unit tests less flaky, but not enough.  I'm still on vacation and will take a closer look when I get back, but my guess is you're getting bit by this.

mmanciop:
Thanks for the merge. I'll prepare some PRs to make sure that there is only one Future created for any HystrixCommand. I think that a little more checks in that class will prevent obscure when users get too creative :-)

mmanciop:
This feature is unfortunately not yet documented in the Wiki. I have created #31377 to track it.

mattrjacobs:
Fixed #1377 with a Wiki update [here](https://github.com/Netflix/Hystrix/wiki/Configuration#executionisolationthreadinterruptoncancel)

PR #1271, opened 2016-07-08
Fix race condition in HystrixThreadPoolKey.asKey method

daniilguit:
Fix rare case when this method would return different values for the same key

The problem would happen when:

1. Thread1: No such key when intern.get(name), so k was null
2. Thread1: We would enter if-statement
3. Thread2: Other thread would add key to intern map
4. Thread1: We will not put new key to intern map, but we will return newly created value instead of one stored in map

### DIFF
--- a/hystrix-core/build.gradle
+++ b/hystrix-core/build.gradle
@@ -5,7 +5,7 @@ dependencies {
     compile 'com.netflix.archaius:archaius-core:0.4.1'
     compile 'io.reactivex:rxjava:1.1.5'
     compile 'org.slf4j:slf4j-api:1.7.0'
-	compile 'org.hdrhistogram:HdrHistogram:2.1.7'
+    compile 'org.hdrhistogram:HdrHistogram:2.1.7'
     testCompile 'junit:junit-dep:4.10'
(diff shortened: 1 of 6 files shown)


chrisgray:
Should we make similar fixes to [HystrixCommandGroupKey](https://github.com/Netflix/Hystrix/blob/master/hystrix-core/src/main/java/com/netflix/hystrix/HystrixCommandGroupKey.java#L50-L56) and [HystrixCommandKey](https://github.com/Netflix/Hystrix/blob/master/hystrix-core/src/main/java/com/netflix/hystrix/HystrixCommandKey.java#L48-L55)?

daniilguit:
I think it worth to extract this functionality to new utility class.

mattrjacobs:
This looks like a good change to me.  Thanks @daniilguit ! Would you like me to merge this or wait until you're done with your refactoring?

daniilguit:
Let's merge the final result, I will update pull request in a few days.

daniilguit:
Committed code unification.

mattrjacobs:
Thanks @daniilguit !  Looks great.

PR #1274, opened 2016-07-10
Make toObservable lazy.  

mattrjacobs:
Prevents work (like cache key calculation) from happening before subscription

* This also implies that queue()/observe() will not throw anymore
* The IllegalStateException now gets wrapped in a HystrixRuntimeException


NiteshKant:
I am not familiar with the code but conceptually wrapping it in `.defer()` makes sense for lazy method `toObservable()`

PR #1279, opened 2016-07-14
Allow scalar commands (HystrixCommand) to set completion state on an onNext

mattrjacobs:
* This allows unsubscribes that happen between onNext/onCompleted to not affect the internal Hystrix command state

* That situation arises naturally fairly frequently - examples include a downstream take(1) or Observable.zip()


mattrjacobs:
Replaced by #1300

PR #1294, opened 2016-07-29
Make sure thread-pools actually shutdown

thesmith:
When HystrixThreadPool.shutdown is called with a timeout, a single call to awaitTermination is called.
This is now done in a loop to make sure the thread-pool is definitely shutdown.


thesmith:
Hmmm.. that passes for me locally.

mattrjacobs:
I made an unrelated unit test that was intermittently failing less flaky, and Travis worked this time.  Apologies for that.

Thanks for the commit

PR #1320, opened 2016-08-12
Javanica: basic Observable Collapser support

zsoltm:


mattrjacobs:
Thanks for the contribution @zsoltm !  Are you able to include any unit tests in this PR that demonstrate how to use the Observable-Collapsers in Javanica?

hystrix-contrib/hystrix-javanica/src/main/java/com/netflix/hystrix/contrib/javanica/aop/aspectj/HystrixCommandAspect.java
```diff
@@ -168,33 +167,38 @@ public MetaHolder create(Object proxy, Method collapserMethod, Object obj, Objec
...
-            if (batchCommandMethod == null || !batchCommandMethod.getReturnType().equals(List.class)) {
+
+            if (batchCommandMethod == null)
+                throw new IllegalStateException("batch method is absent: " + hystrixCollapser.batchMethod());
+
+            Class<?> batchReturnType = batchCommandMethod.getReturnType();
+            Class<?> collapserReturnType = collapserMethod.getReturnType();
+            boolean observable = collapserReturnType.equals(Observable.class);
+
+            if (!batchReturnType.equals(List.class))
```
dmgcodevil: I guess this check is redundant because
 ```getDeclaredMethod(obj.getClass(), hystrixCollapser.batchMethod(), List.class)``` already checks return type, if batch method return type isn't List then ```getDeclaredMethod``` returns null.

hystrix-contrib/hystrix-javanica/src/main/java/com/netflix/hystrix/contrib/javanica/aop/aspectj/HystrixCommandAspect.java
```diff
@@ -260,14 +265,26 @@ private static Method getAjcMethodFromTarget(JoinPoint joinPoint) {
...
+    }
+
+    private static Class<?> getFirstGenericParameter(final Type type, final int nestedDepth) {
+        int cDepth = 0;
+        Type tType = type;
+
+        for (int cDept = 0; cDept < nestedDepth; cDept++) {
+            if (!(tType instanceof ParameterizedType))
+                throw new IllegalStateException(String.format("Sub type at nesting level %d of %s is expected to be generic", cDepth, type));
+            tType = ((ParameterizedType) tType).getActualTypeArguments()[0];
```
dmgcodevil: I think it should be: ``` tType = ((ParameterizedType) tType).getActualTypeArguments()[cDept] ```
BTW: In a separate branch I'm working on strategic solution to compare generic types. Will refactor this code to use the common feature.

hystrix-contrib/hystrix-javanica/src/main/java/com/netflix/hystrix/contrib/javanica/command/AbstractHystrixCommand.java
```diff
@@ -140,8 +139,8 @@ boolean isIgnorable(Throwable throwable) {
      * @param action the action
      * @return result of command action execution
      */
-    Object process(Action action) throws Exception {
-        Object result;
+    <ReturnType> ReturnType process(Action<ReturnType> action) throws Exception {
```
dmgcodevil: Why do we need generic type here ?

dmgcodevil:
@zsoltm I've reviewed PR and left some comments. Also could you please update README with few examples that demonstrate new functionality. Thanks.

hystrix-contrib/hystrix-javanica/src/main/java/com/netflix/hystrix/contrib/javanica/aop/aspectj/HystrixCommandAspect.java
dmgcodevil: it's legacy code. anyway it's better to remove it

zsoltm:
@dmgcodevil thanks for the review, I made the changes in a hope that this is something useful. I just needed this in my project, hacked it together, and created a PR. I'm not 100% happy the way it ended up, so if you're working on a more elegant solution, feel free to drop this.

dmgcodevil:
@zsoltm PR looks good from my point of view. We should try to reuse existing abstractions rather than adding new once where it's possible.

mattrjacobs:
Thanks for the contribution @zsoltm !  

Also, thanks @dmgcodevil for the design review

PR #1324, opened 2016-08-18
Update readme to address jdk7+

caarlos0:
Seems that hystrix is not supported on jdk6 anymore, so maybe this change will help new users...

Feel free to ignore if not :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ mvn -f download-hystrix-pom.xml dependency:copy-dependencies
 
 It will download hystrix-core-*.jar and its dependencies into ./target/dependency/.
 
-You need Java 6 or later.
+You need Java 7 or later.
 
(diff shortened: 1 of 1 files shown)


mattrjacobs:
AFAIK, setting SOURCE/TARGET compatibility to Java 1.6 in build.gradle should be working to generate JDK6 bytecode.  We can't compile with JDK6 though, because The Nebula plugins only support JDK7+:

> * What went wrong:
java.lang.UnsupportedClassVersionError: nebula/plugin/netflixossproject/NetflixOssProjectPlugin : Unsupported major.minor version 51.0
> nebula/plugin/netflixossproject/NetflixOssProjectPlugin : Unsupported major.minor version 51.0

caarlos0:
hmm, ok

PR #1356, opened 2016-09-20
Update PlatformSpecific to only detect App Engine Standard Environment

### LINKED ISSUE #1352: Hystrix doesn't work with App Engine Flexible Environment
We've been using Hystrix in Google App Engine, and have run into a major issue with the new Flexible Environment in App Engine. Basically, when a new service spins up with a Hystrix client, the first request via Hystrix succeeds and then every request after that fails with `java.lang.IllegalStateException: Cannot create new threads after the request thread stops.`

The GAE `VmRequestThreadFactory` appears to be designed to be used only for a single request, but Hystrix keeps the TheadFactory around much longer. So the first Hystrix request goes through, and once the request is done GAE sets `allowNewRequestThreadCreation = false` on the `VmRequestThreadFactory` from that request. GAE spins up a new `VmRequestThreadFactory` for the next request, but unfortunately Hystrix is stuck with a reference to the `VmRequestThreadFactory` from the first request, which it will keep forever. Once that first request is done, GAE won't allow it to spawn any new threads, resulting in the behavior we're seeing.

I locally patched `PlatformSpecific.isAppEngine()` to always return false, and the issue went away completely. In the Flexible Environment we are no longer subject to the GAE sandbox, so there's no reason to use the custom GAE ThreadFactory when the one from `java.util.current` will work just fine.

I am working on a way to programmatically distinguish the GAE Standard Environment from the Flexible Environment so that Flex users can get the standard ThreadFactory. Once I get that worked out I am happy to throw up a PR that will fix this issue.

cadef:
Fixes #1352.

App Engine now has two environments, Standard and Flexible. The Flexible environment is just a Docker container, and as such can use standard Java threading. This change detects the Flexible environment via an environment variable that does not exist in the Standard environment, but exists for Flexible (both in local development and in production).

### DIFF
--- a/hystrix-contrib/hystrix-metrics-event-stream/src/main/java/com/netflix/hystrix/contrib/metrics/eventstream/HystrixMetricsPoller.java
+++ b/hystrix-contrib/hystrix-metrics-event-stream/src/main/java/com/netflix/hystrix/contrib/metrics/eventstream/HystrixMetricsPoller.java
@@ -80,7 +80,7 @@ public class HystrixMetricsPoller {
         this.listener = listener;
 
         ThreadFactory threadFactory = null;
-        if (!PlatformSpecific.isAppEngine()) {
+        if (!PlatformSpecific.isAppEngineStandardEnvironment()) {
             threadFactory = new MetricsPollerThreadFactory();
         } else {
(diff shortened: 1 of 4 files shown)


mattrjacobs:
Looks great @cadef!  Thanks for the contribution

PR #1376, opened 2016-10-07
Remove outdated comments regarding thread keep alive.

bltb:
@mattrjacobs , what do you think?


mattrjacobs:
Yes, thank you!

bltb:
Oh, by the way... for thread keep alive time, in the code, there are a couple of references to `TimeUnit.MINUTES` and in other places it is possible to pass the unit... do we need to make these consistent?

PR #1382, opened 2016-10-14
Bump JMH to the latest version(1.15)

monkey-mas:
It'd be nice to upgrade JMH to the latest version. :)


mattrjacobs:
Thanks @monkey-mas

PR #1385, opened 2016-10-16
Bump JMH gradle plugin to the latest version(0.3.1)

monkey-mas:
Let's upgrade JMH plugin as well if it's okay :)

ref. https://plugins.gradle.org/plugin/me.champeau.gradle.jmh
ref. https://github.com/Netflix/Hystrix/pull/1382


mattrjacobs:
Thanks!

Issue #1381, opened 2016-10-12
Hystrix Javanica Unwraps and Raises Checked Exceptions


56quarters:
Hi,

Starting with hystrix-javanica 1.5.4 (up to the latest master) it seems that `HystrixCommandAspect` starts unconditionally unwrapping `HystrixRuntimeExceptions` and re-raising the `cause`, no matter what. This becomes a problem when Hystrix commands time out. In this case the `cause` will be an instance of `TimeoutException` which is a checked exception. Thus callers of a method that times out get an `UndeclaredThrowableException` instead of a `HystrixRuntimeException`.

I've created a [demo project](https://github.com/56quarters/hystrix-demo) with a `@HystrixCommand` that always times out to demonstrate. This is a Spring Boot project but the problem is unrelated to Spring Boot.

## Steps

```
git clone https://github.com/56quarters/hystrix-demo.git
cd hystrix-demo
./mvnw test
```

## Expected behavior

The method call raises a `HystrixRuntimeException` with `failureType == TIMEOUT` and `cause` of `TimeoutException`

## Actual behavior

The method call raises an `UndeclaredThrowableException` with cause of `TimeoutException`

This seems to be related to #1285 and #1344 and caused by the changes in 155d4e97a2d019238c939fd074feb9443e09ab68

I *think* the fix for this is something to the effect of changing `HystrixCommandAspect.getCauseOrDefault` to look like the following:

```java
    private Throwable getCauseOrDefault(HystrixRuntimeException e, RuntimeException defaultException) {
        if (e.getCause() == null) return defaultException;
        if (e.getCause() instanceof CommandActionExecutionException) {
            CommandActionExecutionException commandActionExecutionException = (CommandActionExecutionException) e.getCause();
            return Optional.fromNullable(commandActionExecutionException.getCause()).or(defaultException);
        }
        if (e.getFailureType() != FailureType.COMMAND_EXCEPTION) return defaultException;
        return e.getCause();
    }
```

Note the addition of the final `if` to only reraise the cause when the failure is from an exception thrown by the wrapped method.

Thanks.

/cc @hroongta

mattrjacobs:
@dmgcodevil an you take a look at this one and recommend a path forward?

dmgcodevil:
[...]

PR #1389, opened 2016-10-19
iss1344: send fallback exception to client instead of primary command

dmgcodevil:


56quarters:
This works for my use case in #1381 :+1:

dmgcodevil:
I need to update docs as well.

mukteshkrmishra:
@dmgcodevil : changes looks good to me. 👍

mattrjacobs:
Looks good to me too, thanks @dmgcodevil for jumping on it

PR #1393, opened 2016-10-19
iss:1387: Performance improvement in EnvUtils. Read "weavingMode" system property only one time.

ricardoletgo:
Performance improvement in EnvUtils. Read "weavingMode" system property only one time.


mattrjacobs:
Thanks @ricardoletgo!

Issue #1390, opened 2016-10-19
HystrixPlugins.getInstance() returns private instance


sahlex:
In Hystrix 1.5.5, when calling `HystrixPlugins.getInstance()` I get a 
```
java.lang.NoClassDefFoundError: Could not initialize class com.netflix.hystrix.strategy.HystrixPlugins$LazyHolder
```
exception in an OSGi environment.

In 1.4.x it worked without problems.
How is intended to work?

The private inner class effectively prevents the OSGi classloader to make it available to client bundles. As soon as I patch it to be publicly available, I can use the plugin mechanism again. As it worked fine in 1.4.x I'd consider this a defect.

sahlex:
Right currently We are relying on Hystrix 1.5.6 so I provided a patch for this version. I guess the master is on 2.x already. The pull request is provided for the master - so myabe a backport to 1.5.7 is necessary?
Anywasy - thanks for the collaboration!

agentgt:
Hmm that is bizarre. Are you sure it is OSGI because if Hystrix can't be initialized you will get that error on subsequent calls. 

> _In most JVM implementations, if construction of Something fails, subsequent attempts to initialize it from the same class-loader will result in a NoClassDefFoundError failure._ (see https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom).

Which makes me concerned that OSGi is calling it more than once.

However because the error of `NoClassDefFoundError` is pretty bad regardless of OSGi perhaps the standard [double checked locking](https://en.wikipedia.org/wiki/Double-checked_locking) would be a better solution.

Thus I  don't recommend @sahlex  changes which will add yet another thing that is `public` (and thus effectively changing the API) but rather the a double checked locking instantiation. I was in fact going to do it this way original but thought it might be cleaner the lazy holder way.

sahlex:
It turned out that after setting up the environment freshly, the exception does not appear any more.
So it seems that it popped up due to some OSGi class-loader mess.

Sorry for the inconvenience, please close the issue.

Thanks a lot anyways.

Alexander.

wentfar:
I have met the same error:

_"Could not initialize class com.netflix.hystrix.strategy.HystrixPlugins$LazyHolder"_

env:jdk8

how to solve?

wentfar:
I have met the same error:

"Could not initialize class com.netflix.hystrix.strategy.HystrixPlugins$LazyHolder"

env:
[...]

PR #1401, opened 2016-10-24
#1390 HystrixPlugins.getInstance() returns private instance

sahlex:
changed inner class HystrixPlugins$LazyHolder to public.

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/strategy/HystrixPlugins.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/strategy/HystrixPlugins.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2012 Netflix, Inc.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


mattrjacobs:
My understanding is that the inner-class `HystrixPlugins$LazyHolder` must be private in order to provide a guarantee that it is not initialized until explicitly asked for via `HystrixPlugins.getInstance()`.  Do I have that right, @agentgt?

If that's correct, then that is in conflict with how OSGi wants to set things up.  We can start looking at other initialization strategies unless someone has a good idea on how to satisfy both goals.

agentgt:
That is correct. I also commented (rambled on and on 😄 ) on #1390.

I took a look at other OSGi friendly libraries that have singletons (ie slf4j) and it appears the consensus is to go the volatile double checked locking route. Originally I was concerned with the volatile but I figure if slf4j is doing it it probably is ok performance wise. It actually is fairly difficult anyway to test for the performance part.

You probably already know but double locking looks like this (copied from wikipedia):

```
class Foo {
    private volatile Helper helper;
    public Helper getHelper() {
        Helper result = helper;
        if (result == null) {
            synchronized(this) {
                result = helper;
                if (result == null) {
                    helper = result = new Helper();
                }
            }
        }
        return result;
    }
}
```

Replace Helper and Foo with HystrixPlugins.

sahlex:
Thanks for investigating! In the meantime I started to set my server up freshly. Out of whatever reasons, the class loading error does not show up any more when using the originally released hystrix-core 1.5.6 artifact.

So the whole thing turned out to be due to some misconfiguration of my environment. 

Sorry for the inconvenience.

Alexander

sahlex:
superseded

mattrjacobs:
Thanks for chiming in @agentgt, I appreciate it.  Looks like we'll leave the init-strategy alone

PR #1408, opened 2016-10-30
Fix typo in collapser options keyword name (from shared-fn to shard-fn)

crimeminister:
Fixes a typo in the `:hystrix/shard-fn` keyword option to the collapser creation macro.


mattrjacobs:
Thanks @crimeminister !  Nice catch

crimeminister:
Thanks for the useful library!

PR #1397, opened 2016-10-20
Add option to raise HystrixRuntimeException

michaelcowan:
Adds the option to allow _HystrixRuntimeException_ to be thrown by `HystrixCommandAspect`.

The new property _raiseHystrixExceptions_ is supported by both `DefaultProperties` and `HystrixCommand`.

e.g.
```java
@DefaultProperties(
    ignoreExceptions = BadRequestException.class,
    raiseHystrixExceptions = {HystrixException.RUNTIME_EXCEPTION})
```

This pattern is extensible should people wish to add support for other Hystrix exceptions.

PR [requested](https://github.com/Netflix/Hystrix/issues/1344#issuecomment-254999908) by @dmgcodevil

Issue #1344, opened 2016-09-02
Javanica: Option to send fallback exception to client instead of primary command


nytins:
We are having an issue because javanica always sends the primary command exception to the client.

**Sample code**
```
@HystrixCommand(ignoreExceptions = {VallidationException.class}, fallbackMethod="getUserEmailsFallback")
List<String> getUserEmails(String userId) {
  return callFirstService(userId);
}

@HystrixCommand(ignoreExceptions = {VallidationException.class})
List<String> getUserEmailsFallback(String userId) {
  return callSecondService(userId);
}
```
**What is happening**
If the first service is down the first command throws `ConnectException` and falls back to fallback method.
If the fallback method throws `ValidationException` then the original `ConnectException` is thrown back to client.

This behaviour is undesired because its just a `ValidationException` and ideally we should be sending a message to client saying there is some validation exception.

**What will make it better**
If fallback throws any of the ignored exceptions then propagate that exception to the client instead of the exception from the primary method.

mukteshkrmishra:
Which version of Javanica you are using?

nytins:
`1.4.21`

dmgcodevil:
@nytins as it was discussed before it's hard to justify importance of an exception thrown from fallback over one thrown from command. Some clients are interested only in command exception, having said that I think it should be configurable.

nytins:
@dmgcodevil here the importance of the fallback exception is that it is one of the ignored exceptions.
Ie, primary failed due to actual backend issues, fallback failed only because of invalid user input (which is an ignored exception). 
In this case doesn't it make sense to send the invalid input exception to client?

nytins:
Another thing I noticed is that the behavior is different when the fallback method is async.

- Sync command, sync fallback --> command exception ([sample code](https://github.com/nytins/hystrix-example/blob/master/src/test/java/com/nytins/hystrix/fallback/SyncCommandSyncFallbackTest.java))
- Async command, sync fallback --> command exception ([sample code](https://github.com/nytins/hystrix-example/blob/master/src/test/java/com/nytins/hystrix/fallback/AsyncCommandSyncFallbackTest.java))
- Async command, async fallback --> fallback exception ([sample code](https://github.com/nytins/hystrix-example/blob/master/src/test/java/com/nytins/hystrix/fallback/AsyncCommandSyncFallbackTest.java))

The below code snippet throws `CommandException`
```
@Autowired
private Service service;

@Test
public void test() throws Throwable {
[...]

PR #1412, opened 2016-11-03
Javanica/Support raiseHystrixExceptions for Observables

michaelcowan:
Adds support for Observable HystrixCommands to optionally raise `HystrixRuntimeException`. 
This extends upon https://github.com/Netflix/Hystrix/pull/1397

e.g.
```java
@HystrixCommand(raiseHystrixExceptions = {HystrixException.RUNTIME_EXCEPTION})
public Observable<User> getUserById(final String id) {
    // ...
}
```

### DIFF
--- a/hystrix-contrib/hystrix-javanica/README.md
+++ b/hystrix-contrib/hystrix-javanica/README.md
@@ -326,9 +326,22 @@ Based on [this](https://github.com/Netflix/Hystrix/wiki/How-To-Use#ErrorPropagat
     }
 ```
 
-If `userResource.getUserById(id);` throws an exception that type is _BadRequestException_ then this exception will be wrapped in ``HystrixBadRequestException`` and re-thrown without triggering fallback logic. You don't need to do it manually, javanica will do it for you under the hood. It is worth noting that a caller will get root cause exception, i.e. user ``BadRequestException``. A caller always gets root cause exception, never ``HystrixBadRequestException`` or ``HystrixRuntimeException`` except the case when executed code explicitly throws those exceptions. 
+If `userResource.getUserById(id);` throws an exception that type is _BadRequestException_ then this exception will be wrapped in ``HystrixBadRequestException`` and re-thrown without triggering fallback logic. You don't need to do it manually, javanica will do it for you under the hood.
 
-*Note*: If command has a fallback then only first exception that trigers fallback logic will be propagated to caller. Example:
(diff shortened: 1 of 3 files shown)


mattrjacobs:
Thanks for the contribution, @michaelcowan !  This looks reasonable to me, but before I merge, I'd love feedback from those who discussed #1344/#1397 to review.

/cc @dmgcodevil @jorgefr @56quarters @nytins @mukteshkrmishra @hroongta

dmgcodevil:
@mattrjacobs you already merged the PR for regular commands, nothing new in this PR. BTW: you merged [PR ](https://github.com/Netflix/Hystrix/pull/1397) with pending comments, in next time could you wait until all comments resolved ?

michaelcowan:
@dmgcodevil Thats weird - I don't see any unresolved comments on https://github.com/Netflix/Hystrix/pull/1397. It shows just myself and @mattrjacobs as the only _participants_ on my screen.

dmgcodevil:
@michaelcowan I left few comments for sure, no idea why they aren't visible for you. Probably I did something wrong.

michaelcowan:
@dmgcodevil Is there something I can maybe answer now?

dmgcodevil:
@michaelcowan why do we need array type for new added property? I thought it's either true or false. Is there an opportunity of extending this in the future ? Is global properties working for the new feature?

mukteshkrmishra:
@mattrjacobs : Changes looks good to me.

michaelcowan:
@dmgcodevil I used an array for two reasons, which are actually pretty much your other two questions :)

1. I wanted to be able to default to `null`*, so that I could support the global properties. If I defaulted to a `false` in default properties there would be no way to know if the user explicitly set the value or not.
2. I thought it might be a nice pattern to expand for other Hystrix exception types should it become requested. In particular I was thinking someone might wish to raise `HystrixBadRequestException`. This would allow Javanica to operate in exactly the same way as hand rolling commands.

(*) You can't default to `null` on Object types for some reason
> "Note that null is not a legal element value for any element type." - [JLS](https://docs.oracle.com/javase/specs/jls/se7/html/jls-9.html)

mattrjacobs:
@michaelcowan Thanks for the PR!  Looks good to me.  

@dmgcodevil Likewise, I didn't see any outstanding comments.  Will wait for a +1 from you to merge this one.

dmgcodevil:
@mattjacobs No objections from my side. Although there is a question on
stack overflow that might be related to this, I'm looking into it

mattrjacobs:
Any concerns about merging this is, or is there still more to discuss @dmgcodevil ?

dmgcodevil:
No, go ahead.

mattrjacobs:
Thanks @michaelcowan !

Issue #1283, opened 2016-07-19
Allow (selected) exception to propagate unwrapped


tbvh:
Using a command, and javanica in particular, adds a (hidden) source for `HystrixRuntimeException`s to the code. If the logic that's being wrapped is capable of throwing exceptions, any surrounding catch block is now useless, unless the `HystrixRuntimeException` is unwrapped.

In my case I've used javanica to add hystrix on top of a service layer that can throw 'business' exceptions. I want to see those, so there is no fallback implemented. Now I've wrapped every call to javanica annotated methods to unwrap the `HystrixRuntimeException`.

Instead I would like to be able to 
1. mark a (Runtime)Exception as intended to be thrown unwrapped. 
2. in the case of javanica, any (checked?) exception that the annotated method is declaring it can throw should be thrown unwrapped.

Nb. In both cases I intend to let these exception count against the errors of the command and to potentially trigger the circuit breaker, if one is configured.

To mark RuntimeExceptions I'm thinking along the lines of an interface like `HystrixUnwrappedException`. Any exception that implements this interface, should be rethrown and not be wrapped in `HystrixRuntimeException`, if they are caught by Hystrix.

Is there any interest in such a feature? Maybe if 1 and 2 are considered separately?

mattrjacobs:
Yeah, I can imagine this being useful.  I'd be happy to review a PR for it.  Good idea!

ashwgupt:
+1 for it.

Can it then also enable one to decide upon fallback method name based upon the original (checked) exception?

That can help one to define various fallback paths for a Hystrix protected method to take based upon the error/exception received.

tbvh:
I added a pr that fixes 1, by defining an Interface that can be implemented by exceptions. Those exceptions will be thrown as-is, not wrapped by HystrixRuntimeException. The fallback is not triggered for these exceptions.

I want to extend on this functionality for 2, by overriding decomposeException and shouldNotBeWrapped in the Javanica implementations.

Feedback is welcome.

tbvh:
When I wanted to pick up work on point 2, I noticed two things.

1. The implications of the change are much larger than I originally envisioned. 
    1. The behavior conflict with the behavior of the `ignoreExceptions` annotation argument.
    2. Because runtime exceptions are already unwrapped, basically nothing will get wrapped in the end.
[...]

PR #1414, opened 2016-11-06
Allow exceptions without HystrixRuntimeException wrapper

tbvh:
Introduce NotWrappedByHystrix to mark Exceptions that should not be wrapped in HystrixRunetimeException.

Partial fix for #1283 .


mattrjacobs:
Thanks for the PR @tbvh !  This looks like a good change to me.  I'm trying to think through side-effects of this change.

A) Only users who opt into using this interface get a change in behavior - that's a good thing
B) At runtime, the effect of the change is that executions that return any Exception instance which implements `ExceptionNotWrappedByHystrix` gets the raw exception
C) These exceptions manifest as FAILURE in the `FailureType` enum and metrics

For C), would it make sense to have a new event type (UNWRAPPED_FAILURE) that would make it easier to distinguish this scenario?

Any other users in the community have thoughts on this change?

Sunshow:
So will this PR be merged to master?

tbvh:
I just posted https://github.com/Netflix/Hystrix/issues/1283#issuecomment-267159727. 

Just to clarify, I've introduced this as a partial fix, but it completely solves point 1 of issue #1283. In the comment I propose to drop point 2, which make this PR in fact a complete fix.

Makes me all the more curious to the review. :)

mattrjacobs:
A couple thumbs-up from the community and I agree with the change.  I think it's fine not to special-case this for metrics.  A failure of the wrapped work should be a FAILURE as far as metrics are concerned, whether the exception is passed back directly or now.  If anyone disagrees with this, then I think that' fine to have in a follow-on PR.

PR #1415, opened 2016-11-08
Fix a little spell mistake in comments.

starlight36:
Fix a little spell mistake in comments, "-1 turns if off" should be "-1 turns it off".


mattrjacobs:
Thanks @starlight36 !

PR #1423, opened 2016-11-16
Write to error log actual dynamicMaxPoolSize from properties if maxPo…

diver-in-sky:
…olSize < corePoolSize


mattrjacobs:
Nice catch - thanks @diver-in-sky !

PR #1435, opened 2016-12-06
Properly handle default values for allowMaximumSizeToDivergeFromCoreSize

### LINKED ISSUE #1434: Runtime changes to allowMaximumSizeToDivergeFromCoreSize are ignored
I am configuring my application using Archaius and a local configuration file (set by `-Darchaius.configurationSource.additionalUrls`), and the `hystrix.threadpool.default.allowMaximumSizeToDivergeFromCoreSize` property seems to be ignored by Hystrix - making all my threadpools have a fixed size = `coreSize`.

I have not yet created a test case to support the theory, but by browsing [HystrixThreadPoolProperties.java](https://github.com/Netflix/Hystrix/blob/master/hystrix-core/src/main/java/com/netflix/hystrix/HystrixThreadPoolProperties.java) I believe this happens because `allowMaximumSizeToDivergeFromCoreSize` is set as a `boolean` instead of a `HystrixProperty<Boolean>`, so it is only read once once during the threadpool initialization phase and any further changes are ignored.

ptab:
Properly handle default values for `allowMaximumSizeToDivergeFromCoreSize` when configuring properties using Archaius.

Fixes #1434

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/HystrixThreadPool.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/HystrixThreadPool.java
@@ -174,7 +174,7 @@ public interface HystrixThreadPool {
             this.queueSize = properties.maxQueueSize().get();
             this.queue = concurrencyStrategy.getBlockingQueue(queueSize);
 
-            if (properties.getAllowMaximumSizeToDivergeFromCoreSize()) {
+            if (properties.getAllowMaximumSizeToDivergeFromCoreSize().get()) {
                 this.metrics = HystrixThreadPoolMetrics.getInstance(threadPoolKey,
                         concurrencyStrategy.getThreadPool(threadPoolKey, properties.coreSize(), properties.maximumSize(), properties.keepAliveTimeMinutes(), TimeUnit.MINUTES, queue),
                         properties);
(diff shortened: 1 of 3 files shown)


ptab:
To give a bit more context: the default value for `HystrixThreadPoolProperties.allowMaximumSizeToDivergeFromCoreSize` is `false` instead of `null`, which means
``` java
private static boolean getValueOnce(String propertyPrefix, HystrixThreadPoolKey key, String instanceProperty, boolean builderOverrideValue, boolean defaultValue) {
  return forBoolean()
         .add(propertyPrefix + ".threadpool." + key.name() + "." + instanceProperty, builderOverrideValue)
         .add(propertyPrefix + ".threadpool.default." + instanceProperty, defaultValue)
         .build()
         .get();
}
```
will never try to read the value for `threadpool.default`.

A workaround for the current code is to either configure it programatically using `HystrixThreadPoolProperties.Setter` or adding a configuration for every specific threadpool key.

mattrjacobs:
I just built a jar with this PR and deployed it into an environment with hystrix-1.5.8.  It has a regression - when the property is set to false, the default maximum (10) is getting picked up.  I'm reviewing the code to determine where that happens.

mattrjacobs:
@ptab Can you try out #1447 and see if it works for you?  It builds on top of this PR (#1435)

It fixes 2 issues:
1) Maintain invariant of maximumSize == coreSize if allowMaximumSizeToDivergeFromCoreSize == false
2) Fix Configuration stream so that it also knows about this invariant

ptab:
I am a bit late to the party, but I did run 1.5.9 against my test case and verified that it works as expected. Thanks for accepting it! 👍

mattrjacobs:
Thanks for the contribution!

PR #1439, opened 2016-12-11
Publish the value of the maximumSize property

ptab:
I'm building some custom dashboards where one of the graphs shows how close the current number of active threads is to the configured maximum size (to measure how close I am to run out of capacity). For this reason I need Hystrix to be able to publish the current value of the `maximumSize` property (currently it only publishes the value of `corePoolSize`, which is not useful if you use dynamic threadpool sizes).

### DIFF
--- a/hystrix-contrib/hystrix-codahale-metrics-publisher/src/main/java/com/netflix/hystrix/contrib/codahalemetricspublisher/HystrixCodaHaleMetricsPublisherThreadPool.java
+++ b/hystrix-contrib/hystrix-codahale-metrics-publisher/src/main/java/com/netflix/hystrix/contrib/codahalemetricspublisher/HystrixCodaHaleMetricsPublisherThreadPool.java
@@ -146,6 +146,13 @@ public class HystrixCodaHaleMetricsPublisherThreadPool implements HystrixMetrics
             }
         });
 
+        metricRegistry.register(createMetricName("propertyValue_maximumSize"), new Gauge<Number>() {
+            @Override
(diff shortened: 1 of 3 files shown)


mattrjacobs:
This change looks like the right idea, but I think it would report incorrectly in 2 cases:

1) allowMaximumSizeToDivergeFromCoreSize == false.  In this case, property.maximumSize reports the default max-size (10), regardless of what the coreSize is.  The actual thread pool uses a maximumSize of coreSize.

2) allowMaximumSizeToDivergeFromCoreSize == true and maximumSize < coreSize.  In this case, the internal prefer the safer option of setting maximumSize to coreSize.

In both cases, properties.maximumSize is reporting what the property value is set to, but not what it used (which is quite confusing!)

Right now, the state of this feature is that all consumers of maximumSize (including these) must do a dance to make sure that they perform the 2 checks above.  It's likely better to centralize this logic somewhere so that consumers don't have to worry about this.  My concern is that hiding what the user has set for maximumSize will make the system hard to understand/debug.

Any thoughts on how to address this, @ptab?  My thought as of now would be to create a new method `HystrixThreadPoolProperties#getActualMaximumSize` that looks different from the getters.  This could be used everywhere internally that needs to know what the actual threadpool maximum size is, while still allowing consumers to find out what the configured value for . `maximumSize` is.  I'd love feedback on if this is a good idea.

mattrjacobs:
Forgot to merge this in before #1449.  That PR builds on top of this one.  Thanks for the contribution, @ptab !

PR #1453, opened 2017-01-04
Add Javanica test for NotWrapped checked exception

tbvh:
Added a test to the javanica project to verify #1414; that a check exception is thrown, without being wrapped in runtime exception.


mattrjacobs:
Great - thanks @tbvh !

PR #1469, opened 2017-01-31
Remove possible NPE

petercla:


spencergibb:
Doesn't seem any different to me.

petercla:
Hello,

I am guessing the code is accessed by multiple threads so the call to the first top.get() may return a non null result but that doesn’t meant that the second call to top.get will return a non null result.

Could be that a different thread pop’ed the stack to leave null

mattrjacobs:
Yeah, this looks like a good change to me.  Thanks for the PR @petercla !  

If the internal state is getting mutated by a different thread while `peek` is inflight, then this method is safer.

PR #1489, opened 2017-03-01
Added exporting of rollingMaxConcurrentExecutionCount value.

LuboVarga:
Make exporting of newly added property to dropwizard. Original issue is https://github.com/Netflix/Hystrix/issues/54 and pull request adding given metrics is mentioned here https://github.com/Netflix/Hystrix/pull/667 .


mattrjacobs:
Thanks @LuboVarga !

Issue #1490, opened 2017-03-02
hystrix-metrics-event-stream implementation using JAX-RS instead of Servlet API


justinjose28:
Right now, hystrix-metrics-event-stream module is dependent on Servlet API. But, I think there are applications that depend only on JAX-RS without any Servlet dependency. Does it make sense to add a module here based on JAX-RS which publishes metrics stream? It is fairly easy and I have created one for our application. I can contribute it if this is the right place to keep it.

mattrjacobs:
Sure, that does sound like a reasonable addition.  I'm happy to review a PR for it.  Thanks!

justinjose28:
@mattrjacobs I have opened a [PR](https://github.com/Netflix/Hystrix/pull/1495)for this issue. Can you please review it?

PR #1495, opened 2017-03-14
hystrix-metrics-event-stream implementation using JAX-RS

justinjose28:
This pull request adds a new module hystrix-metrics-event-stream-jaxrs which is based on JAX-RS rather than Servlet API. This module is helpful for applications which depend only on JAX-RS. See this [issue](https://github.com/Netflix/Hystrix/issues/1490)


mattrjacobs:
Thanks @justinjose28 !  Looks good to me.

keyuri22:
is there any implementation available that can write hystrix.stream data to console or log file?

mattrjacobs:
Not at the moment - but it should be possible to write one yourself.

PR #1502, opened 2017-03-18
delete unuseful code in HystrixConcurrencyStrategy

zzzvvvxxxd:
`threadnumber` no need to be protected
and 
local variable `threadFactory` is never used


mattrjacobs:
Thanks @zzzvvvxxxd

PR #1503, opened 2017-03-19
Fixes #1478 : allow cleaner error propagation

### LINKED ISSUE #1458: Incorrectly getting error "Collapser method must have one argument" from CollapserMetaHolderFactory
I am having a strange issue. I have a method which is annotated with HystrixCommand

`
@Override
	@HystrixCommand(fallbackMethod = "getDataFallback",
	        commandProperties = {
            @HystrixProperty(name = "execution.isolation.thread.timeoutInMilliseconds", value = "60000"),
            @HystrixProperty(name="execution.isolation.thread.interruptOnTimeout", value="TRUE"),
            @HystrixProperty(name = "circuitBreaker.requestVolumeThreshold", value = "20"),
            @HystrixProperty(name = "circuitBreaker.sleepWindowInMilliseconds", value = "10000"),
            @HystrixProperty(name = "circuitBreaker.errorThresholdPercentage", value = "50")
            },
            threadPoolProperties  = {
			@HystrixProperty(name = "coreSize", value = "10")
	})

public List<String> getData(String id, Date date) {
   return null;
}
`

it works most of the times, but rarely without changing anything I get the following exception

` java.lang.IllegalStateException: Collapser method must have one argument: public java.util.List getData(String,Date)`

I checked the code of Hystrix and found that the meta holder factory checks if the HystrixCommand annotation is present then it initializes CommandMetaHolderFactory, otherwise CollapserMetaHolderFactory. In my case, I have the HystrixCommand but it still tries to use CollapserMetaHolderFactory which complains about method not having 1 argument. 

This happens rarely and it's possible no one has seen it yet, I was just wondering if there is a recent fix related to this issue. I am using hystrix 1.5.0.

Thanks in advance.

### LINKED ISSUE #1478: Cleaner error propagation in hystrix core v2
I have exactly the same issue as in #1414 (fix works great btw, thx!) but for `HystrixBadRequestException`.

I tried to implement a `HystrixCommandExecutionHook` but all the methods are executed before `decomposeException` so I don't know how to make it work.

Any idea ?

Thank you !

mNantern:
Fixes #1478 
It is now possible to implements "ExceptionNotWrappedByHystrix" to get back the original exception and not an HystrixBadRequestException


mattrjacobs:
Thanks @mNantern !  Looks great

Issue #1403, opened 2016-10-24
Support @HystrixCommand for rx.Single and rx.Completable similar to rx.Observable


piotrturski:
`@HystrixCommand` doesn't work as expected for methods that return `rx.Single` or `rx.Completable`. it would be nice to have them handled in a same way as `rx.Observable`

piotrturski:
`@HystrixCommand` doesn't work as expected for methods that return `rx.Single` or `rx.Completable`. it would be nice to have them handled in a same way as `rx.Observable`

mattrjacobs:
I think this could get addressed at the `hystrix-core` or `hystrix-javanica` layer.  I'm happy to review PRs for either, if anyone is interested in this.

piotrturski:
i'm willing to try if someone can point me to classes that do that for Observable, so i can understand the architecture

mattrjacobs:
In Javanica, the support is here: https://github.com/Netflix/Hystrix/blob/master/hystrix-contrib/hystrix-javanica/src/main/java/com/netflix/hystrix/contrib/javanica/command/GenericObservableCommand.java.

In `hystrix-core`, the class is here: https://github.com/Netflix/Hystrix/blob/master/hystrix-core/src/main/java/com/netflix/hystrix/HystrixObservableCommand.java.

If you're interested in working on the extension to `hystrix-core`, please let me know, so I can help with design considerations.  Thanks!

piotrturski:
of course i'm open to suggestions. after a quick look at the code, i was thinking about adding `.toSingle()` and `.toCompletable()` somewhere in `HystrixCommandAspect` or create new `ExecutionType`s. it would be perfect to just add some post-processing to existing observableCommand, but yet i don't see any quick way of doing it

dmgcodevil:
[...]

PR #1524, opened 2017-04-02
iss1403: Support @HystrixCommand for rx.Single and rx.Completable sim…

dmgcodevil:
Requested by https://github.com/Netflix/Hystrix/issues/1403


mattrjacobs:
Can you add tests showing the case where fallbacks get triggered?  Specifically, I think that a mismatch between execution type and fallback type should be disallowed.  For example, an execution returning a Completable and a fallback returning an Observable is nonsensical.

dmgcodevil:
@mattrjacobs thanks for reviewing. I added more test cases for successful scenarios, reviled some issues and resolved them.  Regarding a 'failure' scenarios, there is only one special that requires a workaround is ```Completable```, ```Completable``` isn't parametrized class but it's still a RX type which is a container in some sense. I added support for RX type because from javanica perspective the following code is correct: 

```java
public static class Service {
        @HystrixCommand(fallbackMethod = "fallback")
        public Observable<User> command() { throw new IllegalStateException(); }
        private User fallback() { return null; }
    }
```
after a little thinking I thought that it makes sense to support this behavior for Completable, so the following code is legit as well:

public static class Service {
        @HystrixCommand(fallbackMethod = "fallback")
        public Completable command() { throw new IllegalStateException(); }
        private User fallback() { return null; }
    }
```
I will update docs shortly

dmgcodevil:
@mattrjacobs I guess you can merge this one

dmgcodevil:
@mattrjacobs 

- [ ] review
- [ ] merge

mattrjacobs:
Sorry for the delay on this @dmgcodevil .  I'm still trying to figure out how much value this adds if `hystrix-core` doesn't contain the equivalent functionality.

My understanding of this feature is that it will take a user-provided Single/Completable and convert that into a `HystrixObservableCommand`.

I'm writing a unit test to check that submitting a `Completable` even works in vanilla Hystrix

mattrjacobs:
Confirmed this in https://github.com/Netflix/Hystrix/pull/1574

mattrjacobs:
So my thought is that this will help users on `hystrix-javanica` and I don't see a reason not to merge.

PR #1527, opened 2017-04-05
Update build.gradle

Psynbiotik:
commons-collections is no longer used in this project, previously it was only used for a null check in FallbackMethod.java


mattrjacobs:
Thanks @Psynbiotik !

PR #1531, opened 2017-04-06
assert the circuitBreakerForceClosed property

lholmquist:
The motivation here is the "propertyValue_circuitBreakerForceOpen" is currently being checked to make sure there is a value.  

It would make sense that the "propertyValue_circuitBreakerForceClosed" value also be checked


mukteshkrmishra:
Looks good to me.

mattrjacobs:
Thanks @lholmquist !

PR #1539, opened 2017-04-14
Fixed a potential bug in metrics-event-stream-jaxrs

justinjose28:
I had defined `currentConcurrentConnections` as a static variable in abstract class by mistake which caused it to be shared between implementations which was wrong. Now, I have moved it to the implementation class. This was the reason tests where failing sometimes.


mattrjacobs:
Good catch - sorry I missed it when I reviewed.  Thanks for the PR!

Issue #1509, opened 2017-03-21
AOP Alliance support for Hystrix-Javanica.


justinjose28:
Right now, javanica supports only aspectj based aop. It would be great if javanica can support [AOP Alliance](http://aopalliance.sourceforge.net)  as it is endorsed by most of the Dependency Injection tools like Spring, HK2, Guice etc. I have had some success in bringing in AOP Alliance support to javanica even though it involves a good deal of refactoring. Basically, classes like MetaHolder, HystrixCommandBuilderFactory etc had to be sub classed. Let me know your thoughts.

davidkarlsen:
AOP Alliance is a bit clunky code-wise IMHO. Be aware that you can use Aspectj just as well with spring and usually it is easier too

justinjose28:
AspectJ is great with Spring. But, if you are locked with other DI frameworks like HK2, Guice etc they don't have out of the box integrations with aspectj. For eg: In our application, we use Jersey and decided to go with their default DI which is HK2 and HK2 doesn't support aspectJ. On the other hand AOP Alliance is supported by all major DI frameworks. 

Even if javanica does not provide support for other aop libraries out of the box, it would be great if the available apis can be extended so that third party integrations can be done by interested parties without much hassles. But, the current design is very much limited to AspectJ.

mattrjacobs:
@dmgcodevil Any thoughts on if this is something you are interested in reviewing a PR for/writing yourself?

dmgcodevil:
Hi @justinjose28, I know that the current implementation is far from perfect . Since that moment I learned many things and tried different approaches in software development. Those days I was possessed by infamous software patterns in java: builder, factory , builder factory and such like. Now, I hate it and have been considering the idea of rewriting javanica from scratch for a long while. I'm not adept of refactoring projects with broken design: it's hard, expensive and time consuming. I'd to collaborate with people, share my experience and build solid product which everyone will be happy to contribute. 
Regarding your question: yes, it would be better to support all possible variations of AOP frameworks and tools. Do you have any ideas of how it can be accomplished in general ?  I'm wondering if there are such libraries already.

justinjose28:
Hi @dmgcodevil, Thanks for your reply!

I think current implementation is very solid except that it is not very extendable because of the builder, static methods etc. If there is a plan to rewrite javanica, I am also interested to contribute.

[...]

PR #1544, opened 2017-04-20
AOP Alliance support for Hystrix-Javanica

justinjose28:
See [request](https://github.com/Netflix/Hystrix/issues/1509) for more details.


justinjose28:
One of the Hystrix Core test failed even though I didn't modify anything in it. Is there a way I can do a rebuild/test?

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/command/closure/BaseClosureFactory.java
```diff
@@ -60,27 +33,29 @@ public Closure createClosure(MetaHolder metaHolder, Method method, Object o, Obj
...
+   protected Closure createClosure(String rootMethodName, final Object closureObj) throws Exception {
         if (!isClosureCommand(closureObj)) {
             throw new RuntimeException(format(ERROR_TYPE_MESSAGE, rootMethodName,
                     getClosureCommandType().getName()).getMessage());
         }
         Method closureMethod = closureObj.getClass().getMethod(INVOKE_METHOD);
         return new Closure(closureMethod, closureObj);
     }
+    
+    protected abstract Object execute(T metaHolder, Method method, Object o, Object... args);
```
dmgcodevil: `execute` in Factory ? Factory is responsible only for constructing objects, ideally of course. What's the reasons of having `execute` here ?

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/command/MetaHolder.java
```diff
@@ -15,96 +15,94 @@
...
-        this.fallback = builder.fallback;
-        this.extendedParentFallback = builder.extendedParentFallback;
-        this.observable = builder.observable;
-        this.observableExecutionMode = builder.observableExecutionMode;
-    }
-
-    public static Builder builder() {
-        return new Builder();
-    }
+public abstract class MetaHolder<M extends MetaHolder<M,B>,B extends MetaHolder.Builder<M, B>> {
```
justinjose28: Type Parameters are used at two places in MetaHolder.

1) Constructor
```java
protected  MetaHolder(B builder) {
	this.hystrixCommand = builder.hystrixCommand;
        // other properties here
}
```
2) Copy method

```java
protected abstract B copy();
```
Class level Type parameters can be replaced by method level parameters as shown below.

```java

protected <M extends MetaHolder, B extends MetaHolder.Builder<M, B>> MetaHolder(B builder) {
         this.hystrixCommand = builder.hystrixCommand;
        // other properties here
}
protected abstract <M extends MetaHolder, V extends MetaHolder.Builder<M, V>> V copy();
```
The only down side is that above change will show an `unchecked warning` in classes which implement copy() method. I am not sure which approach(method/class level) is ideal in this case. Let me know your thoughts. `copy()` method is used in `LazyCommandExecutionAction`. Also, do you think there is a better way to handle `copy` operation?

In the builder also, we need to pass `Type` of associated MetaHolder  because following snippet in builder is called from a lot of places and without type info we may not get desired behavior. 

```java
public abstract T build();
```

Replacing above snippet as `public  MetaHolder build() ` will not help as we need to know the exact sub class of MetaHolder that is built at the compile time.

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/aop/AbstractHystrixCommandAspect.java
```diff
@@ -0,0 +1,86 @@
...
+import com.netflix.hystrix.contrib.javanica.exception.CommandActionExecutionException;
+import com.netflix.hystrix.contrib.javanica.exception.FallbackInvocationException;
+import com.netflix.hystrix.exception.HystrixBadRequestException;
+import com.netflix.hystrix.exception.HystrixRuntimeException;
+
+public abstract class AbstractHystrixCommandAspect<T extends MetaHolder<T, V>, V extends MetaHolder.Builder<T, V>> {
+
+	protected abstract AbstractHystrixCommandBuilderFactory<T,V> getCommandBuilderFactory();
+
+	public Object execute(AbstractCommandMetaHolderBuilder<T, V> metaHolderBuilder) throws Throwable {
```
justinjose28: Yeah, It makes sense to pass `MetaHolder`. Will correct it.

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/aop/AbstractHystrixCommandAspect.java
```diff
@@ -0,0 +1,86 @@
...
+				cause = getCause((HystrixRuntimeException) cause);
+			}
+		} else if (cause instanceof CommandActionExecutionException) { // this situation is possible only if a callee throws an exception which type extends Throwable directly
+			CommandActionExecutionException commandActionExecutionException = (CommandActionExecutionException) cause;
+			cause = commandActionExecutionException.getCause();
+		}
+
+		return Optional.fromNullable(cause).or(e);
+	}
+
```
justinjose28: Will Correct it.

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/aop/AbstractCommandMetaHolderBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+import com.netflix.hystrix.contrib.javanica.annotation.DefaultProperties;
+import com.netflix.hystrix.contrib.javanica.annotation.HystrixCollapser;
+import com.netflix.hystrix.contrib.javanica.annotation.HystrixCommand;
+import com.netflix.hystrix.contrib.javanica.command.ExecutionType;
+import com.netflix.hystrix.contrib.javanica.command.MetaHolder;
+import com.netflix.hystrix.contrib.javanica.utils.AopUtils;
+import com.netflix.hystrix.contrib.javanica.utils.FallbackMethod;
+import com.netflix.hystrix.contrib.javanica.utils.MethodProvider;
+
+public abstract class AbstractCommandMetaHolderBuilder<T extends MetaHolder<T,V>, V extends MetaHolder.Builder<T,V>> {
```
justinjose28: Yeah. It makes sense to rename it to AbstractMetaHolderFactory.

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/aop/AbstractCommandMetaHolderBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+		this.builder = builder;
+		this.method = collapserMethod;
+		this.obj = obj;
+		this.args = args!=null?args:new Object[0];
+		this.proxyClass=proxyClass;
+	}
+
+	public T build() {
+		AbstractCommandMetaHolderBuilder.HystrixPointcutType type = of(method);
+		setFallbackMethod();
```
justinjose28: will rename it.

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/aop/AbstractCommandMetaHolderBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+				builder.defaultGroupKey(defaultProperties.groupKey());
+			}
+			if (StringUtils.isNotBlank(defaultProperties.threadPoolKey())) {
+				builder.defaultThreadPoolKey(defaultProperties.threadPoolKey());
+			}
+		}
+		
+	}
+
+	protected  void setFallbackMethod() {
```
justinjose28: noted :)

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/aop/AbstractCommandMetaHolderBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+	}
+	protected Method getBatchCommandMethod(HystrixCollapser collapser){
+		return getDeclaredMethod(proxyClass, collapser.batchMethod(), List.class);
+	}
+	protected Class<?> getFirstGenericParameter(Type type) {
+		return getFirstGenericParameter(type, 1);
+	}
+
+
+	protected Class<?> getFirstGenericParameter(final Type type, final int nestedDepth) {
```
justinjose28: > Also: why do we need this method at all since it's used only for nestedDepth=1, can we simplify this ?

This method is called from two places one with depth=1 and the other one with depth=0. 

Regarding Accessors, I will make it static or final based on what ever is appropriate.

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/aop/AbstractHystrixCommandBuilderFactory.java
```diff
@@ -1,76 +1,59 @@
...
+import com.netflix.hystrix.contrib.javanica.utils.MethodProvider;
 
-    public HystrixCommandBuilder create(MetaHolder metaHolder) {
+public abstract class AbstractHystrixCommandBuilderFactory<T extends MetaHolder<T,V>,V extends MetaHolder.Builder<T,V>> {
+	private AbstractCacheInvocationContextFactory<T,V> cacheContextFactory;
+	
+	protected AbstractHystrixCommandBuilderFactory(AbstractCacheInvocationContextFactory<T,V> cacheContextFactory){
+		this.cacheContextFactory=cacheContextFactory;
+	}
+
```
justinjose28: noted.

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/aop/AbstractHystrixCacheAspect.java
```diff
@@ -0,0 +1,38 @@
...
+
+import com.netflix.hystrix.contrib.javanica.cache.CacheInvocationContext;
+import com.netflix.hystrix.contrib.javanica.cache.HystrixRequestCacheManager;
+import com.netflix.hystrix.contrib.javanica.cache.annotation.CacheRemove;
+import com.netflix.hystrix.contrib.javanica.command.ExecutionType;
+import com.netflix.hystrix.contrib.javanica.command.MetaHolder;
+
+public abstract class AbstractHystrixCacheAspect<T extends MetaHolder<T, V>, V extends MetaHolder.Builder<T, V>> {
+
+	public void execute(CacheMetaHolderBuilder<T, V> metaHolderBuilder) {
```
justinjose28: Yes. will correct it.

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/aop/AbstractHystrixCacheAspect.java
```diff
@@ -0,0 +1,38 @@
...
+		CacheInvocationContext<CacheRemove> context = getCacheFactory().createCacheRemoveInvocationContext(metaHolderBuilder.build());
+		HystrixRequestCacheManager.getInstance().clearCache(context);
+	}
+
+	protected abstract AbstractCacheInvocationContextFactory<T,V> getCacheFactory();
+
+	public static abstract class CacheMetaHolderBuilder<T extends MetaHolder<T, V>, V extends MetaHolder.Builder<T, V>> {
+		protected final Method method;
```
justinjose28: noted

hystrix-contrib/hystrix-javanica/hystrix-javanica-aspectj/build.gradle
```diff
@@ -88,32 +90,24 @@ ext {
...
 
 dependencies {
     compileApi project(':hystrix-core')
-    ajtools "org.aspectj:aspectjtools:$aspectjVersion"
-    testRuntime "org.aspectj:aspectjrt:$aspectjVersion"
-    compileApi "org.aspectj:aspectjweaver:$aspectjVersion"
+	compileApi project(':hystrix-javanica-core')
+	compileApi "org.aspectj:aspectjweaver:$aspectjVersion"
     compile "org.aspectj:aspectjrt:$aspectjVersion"
+    ajtools "org.aspectj:aspectjtools:$aspectjVersion"
```
justinjose28: can you please explain what needs to be done here?

hystrix-contrib/hystrix-javanica/hystrix-javanica-aspectj/src/main/java/com/netflix/hystrix/contrib/javanica/aop/aspectj/BeanFactory.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.netflix.hystrix.contrib.javanica.aop.AbstractCacheInvocationContextFactory;
+import com.netflix.hystrix.contrib.javanica.aop.AbstractHystrixCommandBuilderFactory;
+
+
+/**
+ * @author justinjose28
+ *
+ */
+public class BeanFactory {
```
justinjose28: OK

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/command/MetaHolder.java
dmgcodevil: I still don't understand why MetaHolder should be generic ? And why we need to know the concrete type at compile time ? In constructor we you can just use Builder which is super class for all meta holder builders. we can add `copy` in metaholder that also returns super `MetaHolder.Builder`, right ?

hystrix-contrib/hystrix-javanica/hystrix-javanica-aspectj/src/main/java/com/netflix/hystrix/contrib/javanica/aop/aspectj/MethodExecutionAction.java
```diff
@@ -39,17 +41,17 @@
     private final Object object;
     private final Method method;
     private final Object[] _args;
-    private final MetaHolder metaHolder;
+    private final AspectjMetaHolder metaHolder;
```
justinjose28: I think we cannot keep `MethodExecutionAction` in core because all `execute` methods are different for aspectj and aop-alliance.

hystrix-contrib/hystrix-javanica/hystrix-javanica-aspectj/build.gradle
dmgcodevil: basically code formatting

hystrix-contrib/hystrix-javanica/hystrix-javanica-aspectj/src/main/java/com/netflix/hystrix/contrib/javanica/aop/aspectj/MethodExecutionAction.java
dmgcodevil: can be overridable ?

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/command/MetaHolder.java
justinjose28: There are two parts to this question. 

1) Whether MetaHolder should be generic?  
    Need not be. MetaHolder was made generic because of `constructor` and `copy` method. As you mentioned, the same behavior can be accomplished by  passing `MetaHolder.Builder` to constructor and returning `MetaHolder.Builder` from `copy` method. The only down side here is compiler warning about  unchecked conversion.

2) Whether MetaHolder.Builder should be generic?
    It should be generic and the signature should be something like `public abstract static class Builder<M extends MetaHolder, B extends Builder<M, B>>`. I think there is no confusion around `B` as it is required for inheritable builder just like you mentioned in `Foo` example above. My point is that `M` is required for `build` method. `build` method cannot return just MetaHolder. To get a better picture, see how MetaHolder.Builder is used in `AbstractCommandMetaHolderBuilder` which is given below.

```java
public abstract class AbstractCommandMetaHolderBuilder<T extends MetaHolder, V extends MetaHolder.Builder<T,V>> {
    protected final V builder;
    public T build() {
        // Here, If builder returns MetaHolder then we need to cast it to T which is not correct in my opinion.
         return builder.build();
    }
}
```

I hope I have explained it better :) Let me know if I need to explain further.

hystrix-contrib/hystrix-javanica/hystrix-javanica-aspectj/src/main/java/com/netflix/hystrix/contrib/javanica/aop/aspectj/MethodExecutionAction.java
justinjose28: Basically, they have nothing in common and ideally this is something each implementation(aspectj, aop-alliance etc) should define I think. I feel we will not gain much by providing an abstraction here. Anyway, let me think fresh on this tomorrow morning:)

hystrix-contrib/hystrix-javanica/hystrix-javanica-aspectj/src/main/java/com/netflix/hystrix/contrib/javanica/aop/aspectj/AsyncClosureFactory.java
```diff
@@ -36,15 +37,15 @@ public static AsyncClosureFactory getInstance() {
      * {@inheritDoc}.
      */
     @Override
-    boolean isClosureCommand(Object closureObj) {
+    protected boolean isClosureCommand(Object closureObj) {
```
justinjose28: `AsyncClosureFactory` extends `AbstractClosureFactory` which is aspectj specific.

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/command/closure/BaseClosureFactory.java
justinjose28: Will rename it.. I am bad with names:). Any suggestions?

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/aop/AbstractCommandMetaHolderBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+		if (tType instanceof ParameterizedType)
+			return (Class<?>) ((ParameterizedType) tType).getRawType();
+		else if (tType instanceof Class)
+			return (Class<?>) tType;
+
+		throw new UnsupportedOperationException("Unsupported type " + tType);
+	}
+	public enum HystrixPointcutType {
+		COMMAND, COLLAPSER;
+
```
justinjose28: I use eclipse. It looks like there is no straight forward way to import intellij formatter to eclipse. Let me find some work around.

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/command/MetaHolder.java
dmgcodevil: I understand why we need generics in builder otherwise inheritance wont work. In `build` method you can use specific return type, for instance in my example: `Bar.build()` method returns `Bar`. Specific modules should use specific builders, so there shouldn't be any problems with it., core module should use 'MetaHolder.Buider' and shouldn't be aware of specific builder implementations.

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/command/closure/BaseClosureFactory.java
dmgcodevil: it's not about names, factory shouldn't perform any actions in factory other than creating new objects.

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/command/MetaHolder.java
justinjose28: Agree. So what do you suggest in the example scenario I have given below? **Please note that `AbstractCommandMetaHolderBuilder` is not a subclass of MetaHolder.Builder**. It is another class which accepts `MetaHolder.Builder` as a constructor argument, set default properties and then create a MetaHolder and the one which you suggested to rename to `AbstractMetaHolderFactory`.

```java
public abstract class AbstractCommandMetaHolderBuilder<T extends MetaHolder, V extends MetaHolder.Builder<T,V>> {
    protected final V builder;  // MetaHolder.Builder
    protected AbstractCommandMetaHolderBuilder(V builder){
        this.builder = builder;
    }
    public T build() {
        builder.args(args);
	builder.method(method);
	builder.obj(obj);
        // set other properties here

        // Should I cast to T like 'return (T)builder.build()'
         return builder.build();
    }
}
```

dmgcodevil: Yes, basically you can do cast (isn't so bad):
```java
public abstract class AbstractFactory<T, B extends Foo.Builder<B>> {
    public T create(B builder) {
        return (T) builder.build();
    }
}
```
or without cast

```java
public class Foo {

    private int value;

    public Foo(BasicBuilder builder) {
        value = builder.value;
    }

    interface Buildable<T> {
        T build();
    }

    public static abstract class BasicBuilder<B extends BasicBuilder> {
        private int value = 0;

        public BasicBuilder() {
        }

        public B value(int val) {
            value = val;
            return (B) this;
        }

        abstract Object build();
    }

    public static class Builder extends BasicBuilder<Builder> implements Buildable<Foo> {
        public Foo build() {
            return new Foo(this);
        }
    }

}

/***************************************************************************************/

public class Bar extends Foo {

    private int value;

    protected Bar(Builder builder) {
        super(builder);
    }

    public static class Builder extends Foo.BasicBuilder<Builder> implements Buildable<Bar> {
        private int value;

        public Builder barValue(int value) {
            this.value = value;
            return this;
        }

        public Bar build() {
            return new Bar(this);
        }
    }
}

/***************************************************************************************/

public abstract class AbstractFactory<T, B extends Foo.BasicBuilder<B> & Foo.Buildable<T>> {

    public T create(B builder) {
        return builder.build();
    }

}

/***************************************************************************************/

public class BarFactory extends AbstractFactory<Bar, Bar.Builder> {
}

```

But looks complex to justify one cast, isn't ? your call though

hystrix-contrib/hystrix-javanica/hystrix-javanica-aspectj/src/main/java/com/netflix/hystrix/contrib/javanica/aop/aspectj/MethodExecutionAction.java
dmgcodevil: agree.

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/command/MetaHolder.java
justinjose28: We will go with CAST option. I will try to wrap this up during weekend.

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/command/closure/BaseClosureFactory.java
justinjose28: I think I need to clarify a little bit here. Basically, this class `BaseClousureFactory` does only one thing - create a `Closure` based on the passed method arguments. Basically, it executes the passed `method` and wraps the result of the method execution in a `Closure` object. Here, the execution of method differs based on the implementation(aop-alliance, aspectj etc) and that is the reason I have defined it as a abstract method. So, `execute()` is very much part of `Closure` creation and is not used for any other purpose.

dmgcodevil:
Can we split this into two classes then ?

hystrix-contrib/hystrix-javanica/hystrix-javanica-core/src/main/java/com/netflix/hystrix/contrib/javanica/command/closure/BaseClosureFactory.java
justinjose28: I think it makes sense to remove `ClosureFactory` and its implementations. It is used only by `MethodExecutionAction` and `AopAllianceMethodExecutionAction`. The `execute()` part in `BaseClosureFactory` is already present in these classes for other purposes. So it makes sense to create a common class `AbstractMethodExecutionAction`, keep the closure creation and other common stuffs here and let   `MethodExecutionAction` and `AopAllianceMethodExecutionAction` subclass this class. This is something you already suggested in one of the review comments but I failed to think in this direction.

justinjose28:
@dmgcodevil I have tried to incorporate all your review comments. Since some merge requests were accepted after I submitted PR, I had to resolve those conflicts and rebase with master. Can you please give your feedback and also let me know if I have missed something? 

I am not sure whether I followed the recommended approach for rebase, but all changes that were accepted after my PR are included now.

Travis build is failing in `hystrix-rx-netty-metrics-stream` project which I think has nothing to do with this PR.

Once again I appreciate the valuable time and effort you have spend on reviewing this PR.

justinjose28:
Hi @dmgcodevil , Did you get a chance to go through the latest code after I incorporated review comments?

dmgcodevil:
You need to create a branch: iss1509
don't use master branch

PR #1547, opened 2017-04-25
Support multiple circuit status for Hystrix Dashboard

eunmin:
Hystrix Dashboard displays `[object Object]` on circuit status when isCircuitBreakerOpen is aggregate using Turbine stream.

I fixed it like this :


mattrjacobs:
Nice!  Thanks @eunmin !

mattnelson:
Shouldn't the open count be red to match the existing display?

https://github.com/Netflix/Hystrix/wiki/Dashboard

eunmin:
@mattnelson You're right. I'll fix the color. :)

eunmin:
@mattnelson I have opened a PR(#1551) for this issue.

PR #1556, opened 2017-05-01
bugfix/Fixing typo in fallback error message

Thunderforge:
Error message "failed to processed fallback" should be "failed to process fallback".


mattrjacobs:
Thanks @Thunderforge !

PR #1561, opened 2017-05-02
Add exception-handling to interactions with servlet writers

mattrjacobs:
See https://groups.google.com/forum/#!topic/hystrixoss/4o6VqIW0b7Y

I don't believe implementations of servlet writers should be throwing, but if they do, this code is now more resilient.

### DIFF
--- a/hystrix-contrib/hystrix-metrics-event-stream/src/main/java/com/netflix/hystrix/contrib/sample/stream/HystrixSampleSseServlet.java
+++ b/hystrix-contrib/hystrix-metrics-event-stream/src/main/java/com/netflix/hystrix/contrib/sample/stream/HystrixSampleSseServlet.java
@@ -146,12 +146,16 @@ public abstract class HystrixSampleSseServlet extends HttpServlet {
                             @Override
                             public void onNext(String sampleDataAsString) {
                                 if (sampleDataAsString != null) {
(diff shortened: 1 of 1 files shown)


hystrix-contrib/hystrix-metrics-event-stream/src/main/java/com/netflix/hystrix/contrib/sample/stream/HystrixSampleSseServlet.java
```diff
@@ -146,12 +146,16 @@ public void onError(Throwable e) {
...
-                                    // explicitly check for client disconnect - PrintWriter does not throw exceptions
-                                    if (writer.checkError()) {
+                                    try {
+                                        writer.print("data: " + sampleDataAsString + "\n\n");
+                                        // explicitly check for client disconnect - PrintWriter does not throw exceptions
+                                        if (writer.checkError()) {
+                                            moreDataWillBeSent.set(false);
+                                        }
+                                        writer.flush();
+                                    } catch (Throwable t) {
```
mattnelson: Catching `Throwable` is generally a bad idea, these catches should probably be `Exception`

mattrjacobs:
Thanks for the review, @mattnelson .  Changed per your suggestion

PR #1567, opened 2017-05-07
Handle not wrapped exceptions same way as all other

### LINKED ISSUE #1536: Clarify HystrixBadRequestException and ExceptionNotWrappedByHystrix
This is related to #1503 and #1414 

So my question. Right now exceptions that implement ExceptionNotWrappedByHystrix behave like exceptions wrapped in HystrixBadRequestException. E.g. they do not trigger fallback logic.
For my understanding the goal of #1503 was to unwrap exceptions wrapped in HystrixBadRequestException but not change the logic itself. Something like:

* if it is an exception that implements ExceptionNotWrappedByHystrix and no fallback is awailable, throw it. Otherwise call fallback.

* if it is an exception that implements ExceptionNotWrappedByHystrix wrapped in HystrixBadRequestException throw it without triggering fallback.

Sample:

```java
private static final class HystrixBadRequestExceptionFallback extends HystrixCommand<String> {
    HystrixBadRequestExceptionFallback() {
        super(asKey("hystrix-bad-request-exception-fallback"));
    }

    @Override
    protected String run() throws Exception {
        throw new HystrixBadRequestException("need to fallback", new RuntimeException());
    }

    // NOTE: not executed as expected
    @Override
    protected String getFallback() {
        System.out.println("[HystrixBadRequestExceptionFallback.fallback] should not happen");
        throw new RuntimeException("not expected");
    }
}

private static final class Unwrapped extends RuntimeException implements ExceptionNotWrappedByHystrix {
}

private static final class UnwrappedFallback extends HystrixCommand<String> {
    UnwrappedFallback() {
        super(asKey("unwrapped-exception-fallback"));
    }


    @Override
    protected String run() throws Exception {
        throw new Unwrapped();
    }

    // NOTE: not executed but expected to be
    @Override
    protected String getFallback() {
        System.out.println("[UnwrappedFallback.fallback] should happen???");
        return "UnwrappedFallback fallback";
    }
}

private static class UnwrappedHystrixBadRequestExceptionFallback extends HystrixCommand<String> {
    UnwrappedHystrixBadRequestExceptionFallback() {
        super(asKey("unwrapped-exception-hystrix-bad-request-exception-fallback"));
    }

    @Override
    protected String run() throws Exception {
        throw new HystrixBadRequestException("expected one", new Unwrapped());
    }

    // NOTE: executed but not expected according to HystrixBadRequestException logic
    @Override
    protected String getFallback() {
        System.out.println("[UnwrappedHystrixBadRequestExceptionFallback.fallback] should not happen???");
        throw new RuntimeException("not expected???");
    }
}
```

gagoman:
Closes #1536

First of all I want to note that this change is **breaking** one as far as I can see.

While most of my motivation is in #1536 and I will put an overview and some extra thoughts here:

`Hystrix` had that nice idea that `BadRequestException` was designed to bypass fallbacks. It was a really nice feature that allowed users of `Hystrix` to make a difference between execution errors and some conditional (so called "expected") errors e.g. validation.

#1414 #1503 introduced that nice concept of exception unwrapping which is really cool and I'm grateful to authors to have this feature. It has only one flaw - it has broken an idea of `BadRequestException`. Now any exception that implements `ExceptionNotWrappedByHystrix` leaves no choice to user if he wants to have fallbacks available - not to use it.

This PR solves this issue.
I hope I haven't broken a lot.

So the consequences of this PR:

* users need to review their logic as all fallbacks will start to work in case of exceptions marked by `ExceptionNotWrappedByHystrix`

### DIFF
--- a/hystrix-contrib/hystrix-javanica/src/test/java/com/netflix/hystrix/contrib/javanica/test/common/error/BasicErrorPropagationTest.java
+++ b/hystrix-contrib/hystrix-javanica/src/test/java/com/netflix/hystrix/contrib/javanica/test/common/error/BasicErrorPropagationTest.java
@@ -16,6 +16,7 @@
 package com.netflix.hystrix.contrib.javanica.test.common.error;
 
 import com.netflix.hystrix.HystrixEventType;
+import com.netflix.hystrix.HystrixInvokableInfo;
 import com.netflix.hystrix.HystrixRequestLog;
 import com.netflix.hystrix.contrib.javanica.annotation.HystrixCommand;
 import com.netflix.hystrix.contrib.javanica.annotation.HystrixProperty;
@@ -198,9 +199,9 @@ public abstract class BasicErrorPropagationTest extends BasicHystrixTest {
     }
 
     @Test
-    public void testCommandThrowsNotWrappedException() {
+    public void testCommandWithNotWrappedExceptionAndNoFallback() {
         try {
-            userService.throwNotWrappedCheckedException();
+            userService.throwNotWrappedCheckedExceptionWithoutFallback();
             fail();
         } catch (NotWrappedCheckedException e) {
             // pass
@@ -208,11 +209,26 @@ public abstract class BasicErrorPropagationTest extends BasicHystrixTest {
             fail("'NotWrappedCheckedException' is expected exception.");
         }finally {
             assertEquals(1, HystrixRequestLog.getCurrentRequest().getAllExecutedCommands().size());
-            com.netflix.hystrix.HystrixInvokableInfo getUserCommand = getHystrixCommandByKey("throwNotWrappedCheckedException");
+            HystrixInvokableInfo getUserCommand = getHystrixCommandByKey("throwNotWrappedCheckedExceptionWithoutFallback");
             // record failure in metrics
             assertTrue(getUserCommand.getExecutionEvents().contains(HystrixEventType.FAILURE));
             // and will not trigger fallback logic
             verify(failoverService, never()).activate();
+        }
+    }
+
+    @Test
+    public void testCommandWithNotWrappedExceptionAndFallback() {
+        try {
+            userService.throwNotWrappedCheckedExceptionWithFallback();
+        } catch (NotWrappedCheckedException e) {
+            fail();
+        } finally {
+            assertEquals(1, HystrixRequestLog.getCurrentRequest().getAllExecutedCommands().size());
+            HystrixInvokableInfo getUserCommand = getHystrixCommandByKey("throwNotWrappedCheckedExceptionWithFallback");
+            assertTrue(getUserCommand.getExecutionEvents().contains(HystrixEventType.FAILURE));
+            assertTrue(getUserCommand.getExecutionEvents().contains(HystrixEventType.FALLBACK_SUCCESS));
+            verify(failoverService).activate();
         }
     }
 
@@ -288,8 +304,13 @@ public abstract class BasicErrorPropagationTest extends BasicHystrixTest {
             }
         }
 
+        @HystrixCommand
+        void throwNotWrappedCheckedExceptionWithoutFallback() throws NotWrappedCheckedException {
+            throw new NotWrappedCheckedException();
+        }
+
         @HystrixCommand(fallbackMethod = "voidFallback")
-        void throwNotWrappedCheckedException() throws NotWrappedCheckedException {
+        void throwNotWrappedCheckedExceptionWithFallback() throws NotWrappedCheckedException {
             throw new NotWrappedCheckedException();
         }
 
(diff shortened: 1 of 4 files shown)


mattrjacobs:
Thanks @gagoman .

Let me restate the cases to check that I understand the implications:

* exception thrown neither extends `HystrixBadRequestException` nor implements `ExceptionNotWrappedByHystrix`.  It's eligible for fallback and the exception thrown to caller is wrapped in `HystrixRuntimeException`.
* exception thrown extends `HystrixBadRequestException` but does not implement `ExceptionNotWrappedInHystrix`.  It's not eligible for fallback and the original exception (which extends `HystrixBadRequestException`) is thrown to the caller
* exception thrown implements `ExceptionNotWrappedInHystrix` and does not extend `HystrixBadRequestException`.  It's eligible for fallback and the unwrapped exception is thrown to the caller in the case the fallback is not implements or otherwise fails
* exception thrown both implements `ExceptionNotWrappedInHystrix` and extends `HystrixBadRequestException`.   It's not eligible for fallback and the original exception (which extends from `HystrixBadRequestException`) is thrown to the caller.

Those sound like the correct semantics, and what your PR implements.  The piece which is broken in master is the 3rd case - it does not attempt to fallback.

Did I get that right?

gagoman:
> The piece which is broken in master is the 3rd case - it does not attempt to fallback
Yes, the goal of PR is to fix this.

Answers:

> exception thrown neither extends HystrixBadRequestException nor implements ExceptionNotWrappedByHystrix. It's eligible for fallback and the exception thrown to caller is wrapped in HystrixRuntimeException.

tests are: `testExecutionFailureWithNoFallback` and `testExecutionFailureWithFallbackFailure`

> exception thrown extends HystrixBadRequestException but does not implement ExceptionNotWrappedInHystrix. It's not eligible for fallback and the original exception (which extends HystrixBadRequestException) is thrown to the caller

There are no tests that cover `extends` case. Only available are: `testExecutionHookSemaphoreBadRequestException` and `testExecutionHookSemaphoreBadRequestException`

> exception thrown implements ExceptionNotWrappedInHystrix and does not extend HystrixBadRequestException. It's eligible for fallback and the unwrapped exception is thrown to the caller in the case the fallback is not implements or otherwise fails

tests are: `testNotWrappedExceptionWithFallback` and `testNotWrappedExceptionWithNoFallback`

> exception thrown both implements ExceptionNotWrappedInHystrix and extends HystrixBadRequestException. It's not eligible for fallback and the original exception (which extends from HystrixBadRequestException) is thrown to the caller.

as in #2 there are no tests that cover `extends` case while tests that have exception wrapped in `HystrixBadRequestException` are available:
`testNotWrappedBadRequestWithFallback` and `testNotWrappedBadRequestWithNoFallback`

gagoman:
@mattrjacobs Any update on this? Should I add/change something? Thank you.

mattrjacobs:
Thanks @gagoman !

fastluca:
Hi,
in my project implemented a custom exception `HttpBaseException` which extends `HystrixBadRequestException` and implements `ExceptionNotWrappedByHystrix`. My command throws the `HttpBaseException`, I expected to receive no fallback and the exception propagated, but I received the fallback instead. What I missed to obtain this behaviour? Thank you.

gagoman:
Hello.

I'm not sure about your case. Could you please provide an example?

Is this your case?

```java
private static class BadAndUnwrapped extends HystrixBadRequestException implements ExceptionNotWrappedByHystrix {
    public BadAndUnwrapped(String message) {
        super(message);
    }
}

private static class Sample extends HystrixCommand<String> {

    protected Sample() {
        super(asKey("sample"));
    }

    @Override
    protected String run() throws Exception {
        throw new BadAndUnwrapped("expected one");
    }

    @Override
    protected String getFallback() {
        throw new RuntimeException("not expected");
    }
}

public static void main(String[] args) throws Exception {
    new Sample().run();
}
```

generates next output:

```
Exception in thread "main" hystrix.HystrixSample$BadAndUnwrapped: expected one
	at hystrix.HystrixSample$Sample.run(HystrixSample.java:82)
	at hystrix.HystrixSample.main(HystrixSample.java:92)
```

Hystrix version: **1.5.12**

fastluca:
Hi @gagoman, thank you for your response. Yes, it was exactly my case, I've not tried directly the `run` method, but the `execute` one on `HystrixCommand`.
I've the same Hystrix version: 1.5.12. Now I changed my configuration, with a custom workaround implemented, but calling the `execute` I've the issue described in my previous comment.

gagoman:
You are right about `execute`. Still I can not reproduce, even by using `execute`. Could you please provide an example?

P.S. I would suggest to debug `executeCommandAndObserve` and especially `final Func1<Throwable, Observable<R>> handleFallback` part to see how your `Exception` is treated.

PR #1573, opened 2017-05-09
iss1545: don't get cause from HystrixBadRequestException if command t…

dmgcodevil:
…hrows HystrixBadRequestException without cause


mattrjacobs:
Thanks @dmgcodevil !

PR #1586, opened 2017-05-16
Get CodaHale metrics working by starting up the stream consumers

### LINKED ISSUE #1581: HystrixCodaHaleMetricsPublisher broken in 1.5.x
The metrics publisher seems to be broken in 1.5.x where in 1.4.x it works perfectly.

The following test based on HystrixCodaHaleMetricsPublisherCommandTest *works in 1.4.x* and fails in *1.5.x* 

```java
package com.netflix.hystrix.contrib.codahalemetricspublisher;

import com.codahale.metrics.Gauge;
import com.codahale.metrics.MetricRegistry;
import com.netflix.hystrix.HystrixCommand;
import com.netflix.hystrix.HystrixCommandGroupKey;
import com.netflix.hystrix.HystrixCommandKey;
import com.netflix.hystrix.strategy.HystrixPlugins;
import org.junit.Before;
import org.junit.Test;

import java.util.Map;

import static org.hamcrest.core.Is.is;
import static org.junit.Assert.assertThat;

public class HystrixCodaHaleMetricsPublisherCommandTest {
    private final MetricRegistry metricRegistry = new MetricRegistry();

    @Before
    public void setup() {
        HystrixPlugins.getInstance().registerMetricsPublisher(new HystrixCodaHaleMetricsPublisher(metricRegistry));
    }

    @Test
    public void commandMaxActiveGauge() throws InterruptedException {
        Command command = new Command();
        command.execute();

        for (Map.Entry<String, Gauge> entry : metricRegistry.getGauges().entrySet()) {
            entry.getValue().getValue();
        }


        assertThat((Long) metricRegistry.getGauges().get("test.test.countSuccess").getValue(), is(1L));

    }

    private static class Command extends HystrixCommand<Void> {
        final static HystrixCommandKey hystrixCommandKey = HystrixCommandKey.Factory.asKey("test");
        final static HystrixCommandGroupKey hystrixCommandGroupKey = HystrixCommandGroupKey.Factory.asKey("test");

        Command() {
            super(Setter.withGroupKey(hystrixCommandGroupKey).andCommandKey(hystrixCommandKey));
        }

        @Override
        protected Void run() throws Exception {
            return null;
        }
    }
}
```

mattrjacobs:
Fix #1581 

Issue was twofold:

1) The metric-consuming streams were not being started
2) Since reads are only available after a window passes, we need to wait in the unit test for the value to show up

### DIFF
--- a/hystrix-contrib/hystrix-codahale-metrics-publisher/src/test/java/com/netflix/hystrix/contrib/codahalemetricspublisher/HystrixCodaHaleMetricsPublisherCommandTest.java
+++ b/hystrix-contrib/hystrix-codahale-metrics-publisher/src/test/java/com/netflix/hystrix/contrib/codahalemetricspublisher/HystrixCodaHaleMetricsPublisherCommandTest.java
@@ -12,6 +12,9 @@ import org.junit.Test;
 
 import java.util.Map;
 
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertThat;
+
 public class HystrixCodaHaleMetricsPublisherCommandTest {
     private final MetricRegistry metricRegistry = new MetricRegistry();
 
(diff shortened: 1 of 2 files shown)


hystrix-contrib/hystrix-codahale-metrics-publisher/src/test/java/com/netflix/hystrix/contrib/codahalemetricspublisher/HystrixCodaHaleMetricsPublisherCommandTest.java
```diff
@@ -20,27 +23,28 @@ public void setup() {
...
 
-    @After
-    public void teardown() {
-        HystrixPlugins.reset();
+    @Test
+    public void testCommandSuccess() throws InterruptedException {
+        Command command = new Command();
+        command.execute();
+
+        Thread.sleep(1000);
```
kevlarC: surely you can use await here instead of sleep from https://github.com/awaitility/awaitility
such as   
```
Gauge<Long> successGauge = metricRegistry.getGauges().get("test.test.countSuccess");
await().atMost(1, SECONDS).until(successGauge::getValue, isEqualTo(1L));
```

hystrix-contrib/hystrix-codahale-metrics-publisher/src/test/java/com/netflix/hystrix/contrib/codahalemetricspublisher/HystrixCodaHaleMetricsPublisherCommandTest.java
```diff
@@ -20,27 +23,28 @@ public void setup() {
...
+
+        assertThat((Long) metricRegistry.getGauges().get("test.test.countSuccess").getValue(), is(1L));
```
kevlarC: Hmm ... should this also test the other gauges ?
how about a counter from each stream ?

RollingCommandEventCounterStream
RollingCommandLatencyDistributionStream
RollingCommandUserLatencyDistributionStream
RollingCommandMaxConcurrencyStream

Which stream to timeouts happen on ?

hystrix-contrib/hystrix-codahale-metrics-publisher/src/test/java/com/netflix/hystrix/contrib/codahalemetricspublisher/HystrixCodaHaleMetricsPublisherCommandTest.java
mattrjacobs: I could, but then I introduce an extra dependency for this specific test.  You could argue that a utility like this should be applied across all of the unit tests.  My personal opinions is that the current state isn't enhanced by that much by a wholesale change to this style, but I'm happy to be convinced otherwise

hystrix-contrib/hystrix-codahale-metrics-publisher/src/test/java/com/netflix/hystrix/contrib/codahalemetricspublisher/HystrixCodaHaleMetricsPublisherCommandTest.java
mattrjacobs: Yes, more test coverage here would be great.  Would you like to submit a PR?

PR #1587, opened 2017-05-17
throw IllegalStateException if request cache is not available when clearing

jack-kerouac:
While an `IllegalStateException` is already thrown when the `ConcurrentHashMap` cannot be retrieved for getting or putting to it, this will also throw an exception when it is not available when clearing.


mattrjacobs:
Nice catch.  Thanks @jack-kerouac !

PR #1595, opened 2017-05-29
Possibility to add custom root node for command and thread pool metrics

dstoklosa:
Currently, command and thread pool metrics are stored in the same root ( commandGroup.commandName... and HystrixThreadPool.commadGroup....).

This PR allows to store metrics in the provided root node i.e. hystrix.commandGroup.commandName... , so they are not stored in the same root as other metrics.


mattrjacobs:
LGTM @dstoklosa .  Thanks for the contribution

PR #1604, opened 2017-06-06
Fix metrics ignore CommandGroup when calculating

dmorari:
## Issue :

Metrics in HystrixMetricsPublisherFactory are cached by CommandKey, while the full metric name is generated with CommandGroup being part of the path. 

For example, 
 ```
groupA.commandA
 groupB.commandA
 groupC.commandA
```

Would generate one single metric `groupX.commandA`, where _X_ is the CommandGroup which happened to be executed first. 

Same by-command-key caching is in **HystrixCommandMetrics**. 

Other caching instances are the streams : _RollingCommandEventCounterStream, CumulativeCommandEventCounterStream, CumulativeCommandEventCounterStream,_ etc.. 

## This pull request

The challenging part is with stream registries:  their public API to get instances don't take CommandGroup parameter, making impossible to generate a proper cache key without breaking backward compatibility.  

In order to bypass this limitation, a fake `metricsCommandKey` was introduced, which has name formed using both commandGroup and commandKey.
It is used when instantiating HystrixCommandMetrics instance, as well as in all calls on it.

The HystrixMetricsPublisherFactory takes original commandKey and makes cache key on its own, since later on group+key is used to generate full metric name.

This is a hack, most probably the right solution would be to change the API to take CommandGroup everywhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,25 @@
 # Hystrix Releases #
 
+### Version 1.5.12 ([Maven Central](http://search.maven.org/#search%7Cga%7C1%7Cg%3A%22com.netflix.hystrix%22%20AND%20v%3A%221.5.12%22), [Bintray](https://bintray.com/netflixoss/maven/Hystrix/1.5.12/)) ###
+
+- [Pull 1586](https://github.com/Netflix/Hystrix/pull/1586) Start streams for CodaHale metric consumer, ot get it actually working
+- [Pull 1584](https://github.com/Netflix/Hystrix/pull/1584) Javanica: Wire up allowMaximumSizeToDivergeFromCoreSize thread-pool property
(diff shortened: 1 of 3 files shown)


mattrjacobs:
@dmorari Apologies for the delay - I was out on vacation.  Thanks for the PR.

As you mentioned in the PR, this sort of fix will apply only in the case of the same command key occurring multiple times in the same JVM.  Is this a problem that you run into frequently?  I ask because we have not found ourselves in this position internally.

The unit tests do not pass with this PR.  When I run them locally, the first test that does not pass is `HystrixCodaHaleMetricsPublisherCommandTest.testCommandSuccess`.  I imagine that the metrics publisher is not correctly looking up the metric by name any more.  Can you take a look at that failing test and see what the fix would look like?  Thanks!

PR #1605, opened 2017-06-07
Issue 1603

### LINKED ISSUE #1603: java.lang.RuntimeException: unsubscribed before executing run() with request caching enabled
After upgrading from Hystrix 1.4.x to 1.5.10, we started seeing [this exception](https://github.com/Netflix/Hystrix/blob/fdfb8f970cc42bf29ac8a7438bf40734911e45d5/hystrix-core/src/main/java/com/netflix/hystrix/AbstractCommand.java#L686) from AbstractCommand being logged to our service's output.

This happens when request caching is enabled and the value is retrieved from the cache.  Seemingly it doesn't affect the outcome of the command result and the correct value is returned by `command.queue`.

This seems to be highly timing dependent.  From the investigation I've done so far, I've narrowed it down to the following scenario:

1. Two commands are created on 2 different threads (both having the same Hystrix request context)
2. `command.queue()` is called on them at approximately the same time.
3. The cache key for both commands is the same.
4. One of the commands (`command1`) executes asynchronously.
5. The given exception is thrown from something in `command2` and logged to the console.
6. `command1` returns the result with `command1.isResponseFromCache()` returning false.
7, `command2` returns with the *correct result from the request cache* and `command2.isResponseFromCache()` returns true.

I've been trying (so far unsuccessfully) to reproduce this issue in a unit test.  It seems, however, highly timing dependent.  Any insight that can be provided so that I can produce a unit test that re-creates this scenario would be much appreciated!

atoulme:
Change the observable value returned when the command was unsubscribed from an error to an empty value. This fixes #1603.

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/AbstractCommand.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/AbstractCommand.java
@@ -683,7 +683,7 @@ import java.util.concurrent.atomic.AtomicReference;
                         }
                     } else {
                         //command has already been unsubscribed, so return immediately
-                        return Observable.error(new RuntimeException("unsubscribed before executing run()"));
+                        return Observable.empty();
                     }
                 }
             }).doOnTerminate(new Action0() {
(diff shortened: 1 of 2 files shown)


stevenchurd:
Thanks @atoulme, unfortunately I'm not especially familiar with the Hystrix command state machine.  It certainly seems like it's an invalid assumption that a command never unsubscribes before being run but hopefully someone that knows more can weigh in as well.  @mattrjacobs?

mattrjacobs:
@atoulme Apologies for the delay in getting to this - I was out on vacation.  I agree that your fix is correct.  If the `compareAndSet` returns false, that implies that the command has been unsubscribed, therefore the valid emission is nothing (not an error object as occurs today in master).

Nice find, and thanks for the PR!

PR #1606, opened 2017-06-08
Escape user entered input to avoid HTML injection. This fixes #1456

### LINKED ISSUE #1456: XSS and HTML Injection in title field and stream URL field
imagen

When you click in Add Stream button the injection will be executed:
imagen

You have to click in the injected title or url field to view the JavaSript alert.

atoulme:
Using [underscore escape](http://underscorejs.org/#escape) function, the user entered input is escaped.


mattrjacobs:
Thanks @atoulme !

PR #1624, opened 2017-07-11
Hystrix code corrections

rekhajoshm:
Hystrix code corrections [ Used Hystrix today 👍  , looked a bit into hystrix code ]
- Corrected for NullPointerException case on null command2
- Used current onEmit instead of deprecated onComplete
- Used System.arrayCopy() java method directly
- Used StringBuilder
- inner enum is static by default
- narrowed scope to needed


rekhajoshm:
rebuild this please, error not related to PR.

Issue #1634, opened 2017-07-19
deprecate trivial/duplicated configuration entry metricsRollingPercentileBucketSize


xiaoshuang-lu:

xiaoshuang-lu:
is duplicated by https://github.com/Netflix/Hystrix/pull/1635

PR #1635, opened 2017-07-19
[HYSTRIX-1634] deprecate trivial/duplicated configuration entry metri…

xiaoshuang-lu:
…csRollingPercentileBucketSize


xiaoshuang-lu:
duplicates https://github.com/Netflix/Hystrix/pull/1634

PR #1640, opened 2017-07-24
Fix circuitBreaker may stay closed forever

mskyspet:
HystrixCircuitBreakerImpl use circuitOpen and status to present circuit status.
It is possible 
that circuitOpen == -1L and status == OPEN at the same time which would make attemptExecution return true fore ever.
hystrixcircuitbreakerimpl
In my case, seting sleepWindowInMilliseconds to 999 ms and using a 60% rate fail command is easy to reproduce this issue.

And.
reset should work from inner to outer. 
e.g. a new HystrixCircuitBreakerImpl after HystrixCircuitBreaker.Factory.reset() may still get the old HystrixCommandProperties before HystrixPropertiesFactory.reset().

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/HystrixCircuitBreaker.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/HystrixCircuitBreaker.java
@@ -200,6 +200,8 @@ public interface HystrixCircuitBreaker {
                     });
         }
 
+
+
         @Override
         public void markSuccess() {
             if (status.compareAndSet(Status.HALF_OPEN, Status.CLOSED)) {
@@ -242,7 +244,7 @@ public interface HystrixCircuitBreaker {
             if (properties.circuitBreakerForceClosed().get()) {
                 return true;
             }
-            if (circuitOpened.get() == -1) {
+            if (status.get() == Status.CLOSED) {
                 return true;
             } else {
                 if (status.get().equals(Status.HALF_OPEN)) {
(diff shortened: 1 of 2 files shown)


hystrix-core/src/main/java/com/netflix/hystrix/HystrixCircuitBreaker.java
```diff
@@ -268,7 +270,7 @@ public boolean attemptExecution() {
             if (properties.circuitBreakerForceClosed().get()) {
                 return true;
             }
-            if (circuitOpened.get() == -1) {
+            if (status.get() == Status.CLOSED) {
```
ylgrgyq: maybe it's more consistent to check status with `equals` like :
```
status.get().equals(Status.CLOSED)
```

ylgrgyq:
It seems like the circuit open time will be lost.
The problem lies in the inconsistency between `status` and `circuitOpened`. After thread A change `status` from HALF_OPEN to CLOSED in `markSuccess`, thread B changed `status` from CLOSED to OPEN before thread A had a chance to reset the `HealthCountsStream`. 
Then if thread A runs first, `circuitOpened` will be set to -1 by thread A and set to circuit open time afterwards by thread B. Everything is OK. But if thread B runs first, `circuitOpened` will be set to circuit open time by thread B and set to -1 afterwards by thread A which leaves `status` to OPEN and `circuitOpened` to -1. In the latter scenario, the circuit open time is lost even we check circuit open by it's status in `attemptExcution` and `allowRequest`.

PR #1646, opened 2017-08-01
[WIP] Add OnInterrupt method to handle cleanup of blocking libs

### LINKED ISSUE #1643: Proposal to extend the lifecycle of Hystrix command: Execute cleanup login on timeout
Timeouts in Hystrix interrupt the execute thread (when the execution isolation mode is 'thread' and the 'interruptOnTimeout' is 'true'; but the case per default).

Unfortunately, Java code seldom reacts gracefully and promptly to interruptions: only a handful of Java.lang.Thread methods will actually throw InterruptedException and most of the code out here, in particular the one performing blocking I/O, will happily continue to keep the execution thread busy for the foreseeable future (or even deadlock it when timeouts are not set, e.g., at socket level).

Ideally, one would be able to specify code to execute on timeouts, e.g., invoke a javadoc.sql.Statement.cancel() or a Closeable.close(). This can be done today by spawning another thread in the HystrixCommand.run() method (e.g., via a dedicated executor) and have the latter thread do the blocking work. Hystrix's execution thread joins on the thread doing the blocking work, reacts to the interruption and then "breaks up" the spawned thread as appropriate.

Although this works, it costs a thread (the one doing the blocking work) that could be spared, as the "cleanup on timeout" logic could be executed by the HystrixTimerThread instead. And threads are expensive in terms of memory, by default 1 MB per thread outside of the Heap.

Another option is to do the cleanup in the fallback and rethrow from there the TimeoutException, but it is an even worse hack IMO and violates the (unspoken?) rules of fallback: fast and error proof.

I was thinking of preparing a PR that would:
* Add a 'void onTimeout() throws Exception' overridable method to Hystrix commands (most cleanup operations throw checked exceptions); the onTimeout would be invoked before checking the fallback
* Add a failure type to represent timeout cleanup issues (although I am less sure about whether this is a good idea: failure, e.g., in closing a network connection is very seldom something programs can explicitly react to)

This change is gonna touch a few of the internals, so I would like feedback before I start prototyping it. What says you, oh Hystrix community?

michele-mancioppi:
Concept implementation + tests for an extension of the lifecycles of
Hystrix commands (both imperative and reactive styles) to execute
the AbstractCommand.onInterruption() method when the execution thread
is interrupted.

The goal is to allow graceful cleanup of blocking calls on execution
threads that do not react to interruption, e.g., java.net.Socket.read()
and java.sql.Statement.execute(), but that instead require other methods
to be invoked from a different thread.

In case the interruption is coming from a HystrixTimer thread, the
AbstractCommand.onInterruption() is invoked from the former. In case
of HystrixCommand.queue().cancel(true), the AbstractCommand.onInterruption()
is invoked from the thread invoking the Future.cancel(boolean) method.

TODO: Add hooks

Closes #1643


michele-mancioppi:
It actually dawned on me that we could simply call the new method 'cancel' :-)

michele-mancioppi:
Ping! I'd actually like to have a discussion about this :-)

michele-mancioppi:
Finally the unrelated tests that causes the previous builds to fail have been stabilised, nice :-)

michele-mancioppi:
Almost one year in, still no answer :-)

mmanciop:
More than one year in, still no answer.

PR #1657, opened 2017-08-14
(refactor) some opportunities to use lambda functions

desmorto:


ylgrgyq:
From the README, it seems that Hystrix still wants to support Java 6 which does not have lambda functions yet.

PR #1676, opened 2017-09-11
Add fallbackDisabled metric

haleyw:
Add metric/tagging for indicating when a fallback is skipped because it's disabled.


howardyuan:
LGTM.

PR #1749, opened 2018-01-10
deprecated and move hystrix-dashboard to Netflix-skunkworks

qiangdavidliu:


coffeetocode:
lgtm

vxavictor513:
Why is Hystrix Dashboard deprecated? Is there any alternative?

spencergibb:
Netflix hasn't used it for a long while. They use Atlas

PR #1786, opened 2018-04-09
Stabilize another test

MenschNestor:
The test execution was non-deterministic due to the executor service with two threads. This was executing the runnables with essentially random order and timing. The runnables are now executed one after the other on a single thread, which still allows to test the caching behaviour.

### DIFF
--- a/hystrix-core/src/test/java/com/netflix/hystrix/UnsubscribedTasksRequestCacheTest.java
+++ b/hystrix-core/src/test/java/com/netflix/hystrix/UnsubscribedTasksRequestCacheTest.java
@@ -60,12 +60,6 @@ public class UnsubscribedTasksRequestCacheTest {
         @Override
         protected Boolean run() {
             numOfExecutions.getAndIncrement();
-            try {
-                Thread.sleep(500);
-            }
-            catch (InterruptedException e) {
-                e.printStackTrace();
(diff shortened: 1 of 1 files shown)


MenschNestor:
@howardyuan Thanks for the previous merge. This one fixes a test that failed intermittently, so I did not fix it initially because I could not find the cause at the time.

erichhsun:
@MenschNestor quick question, what method were you using to reproduce the failing tests in question?

MenschNestor:
@erichhsun For this one, I didn't ;-) I've had some experience with using multi-threading in tests.

For the previous ones, the key was finding out that the build VM used by Travis only has 2 cores. This directly impacts the RxJava computation scheduler and the Grizzly web server used in the tests. Once I had found out the system properties I could set to simulate a limited number of cores, the tests were failing consistently.

For example, the property `rx.scheduler.max-computation-threads` allows you to limit the number of threads on the computation scheduler to 2, which mirrors the situation on the build VM.

erichhsun:
@MenschNestor thank you for sharing, that is very helpful.

MenschNestor:
@howardyuan @qiangdavidliu Please merge this PR. If there's any other contact at Netflix that I should be asking instead, please refer me to them.

PR #1796, opened 2018-04-25
Remove cancelled work from the threadpool queue.

wuwen5:
when the ```ScheduledFuture``` is canceled, but the task is has not removed from the threadpool queue. there are very many instance in memory until fullgc.

image
image
image


wuwen5:
#1375

Issue #1807, opened 2018-05-16
Support for marking exception as Business Exception


matelang:
Hi.

We use Hystrix with Feign in a Spring Cloud microservices environment in our startup company, @smartupio.

For each microservice we ship a service-client project which packs the necessary Request/Response value objects, exception types, and a Feign client with a Feign error decoder (https://github.com/coveo/feign-error-decoder).

We find it really odd that there is no way to mark an exception as business exception so that it is handled by Hystrix as a successful outcome. 
We think that It is a fairly common practice in Web Services to return specialized error codes telling the web service client what went wrong. 
In a RESTful context we do achieve this by using standard error codes (4xx, 5xx) enriched with a response body. In our case this is a JSON object containing requestId, timestamp and most importantly an errorCode field indicating what went wrong. E.g. 400 with PasswordValidationException if you try to change your password, or 404 UserNotFoundException if you try to look up a user which does not exist. 

We've learnt the hard way that simply throwing these exceptions enough times in a rolling window will open the Hystrix circuit.

We found that if we extend our exeption from `HystrixBadRequestException` and  `ExceptionNotWrappedByHystrix` then throwing the aforementioned exceptions will not cause the Hystrix circuit to open.

This partially solves the problem since the circuit will not be opened, but it still does not count as a successful request, it's rather ignored completely in statistics.

As a side effect, if a remote web service becomes healthy after the circuit had been closed for technical exceptions/transient errors, but the usage pattern produces more business exceptions than successful HTTP requests (those that return a body) the circuit will stay open. 

There is a fundamental difference between a request (or command in Hystrix terminology) failing because syntactic/semantic client, which is solved great by extending the HystrixBadRequest versus an RPC layer mapping an error message and throwing a business exception in runtime.

We think there is a need for Hystrix to have a marker interface that indicates that a certain exception is actually a success. 
This solution would not be intrusive, it would just enhance the configurability for the use cases where the desired behavior is RPC like, hiding the remoteness of called web service and the method of message passing.

Thanks for taking a time to look into the possibility to do this.

May the source be with you!

PR #1809, opened 2018-05-18
Added support to mark HystrixBadRequestException as metrics (in)transient

matelang:
See issue for proposition: https://github.com/Netflix/Hystrix/issues/1807


mattnelson:
`HystrixBusinessException` sounds like corporate speak and doesn't clearly indicate what is actually wrong. `HystrixBadRequestException` already covers this. I would propose to instead introduce a property that includes `HystrixBadRequestException` with the statistics. That would also allow all current consumers to opt into this feature with only a config change.

matelang:
Hi @mattnelson ,

Thanks for taking the time to look into this.

I am happy to change the name of the exception if it sounds too corporate, unfortunately I can not think of a better name to express the fact that we'd like to throw an exception that indicates that the called method was indeed executed (opposed to the `HystrixBadRequestException` case which, if I understand correctly should translate to something which could not be executed and should be completely ignored from statistics).

Imagine the following example scenario:

```java
public class UserNotFoundException extends HystrixBusinessException{
 //ctors omitted for brevity
}

public interface UserService {
  User findUser(String userName) throws UserNotFoundException;  
}

public class RemoteUserService implements UserService{
  @Override
  public User findUser(String userName){
    ResponseEntity<User> resp = restTemplate.get(...); // get the user
    if (resp.getStatusCode() == 404){   
      throw new UserNotFoundException();
    }

    return resp.getBody();
  }
}
```

I'd expect to treat the user not being found case from the previous example as an absolute success of the operation, because the implementation fulfilled its purpose.
If the above method `findUser` is called from a Hystrix command I'd expect the command to be marked as successful.

I understood that the `HystrixBadRequestException` is more suitable for situations where the operation was skipped or not performed. As an example: for a web service we could skip the actual call if we are unable to serialize an object to be sent to JSON. That means we did not actually call the service, so we should not touch any of the metrics nor the state of the circuit breaker.

I am not sure whether the 2 cases: `HystrixBadRequestException` and `HystrixBusinessException` should be treated as one.

The proposed solution is also non-intrusive. Clients who do not opt in for the BusinessException capability will not see any change in behavior.

mattnelson:
> I'd expect to treat the user not being found case from the previous example as an absolute success of the operation, because the implementation fulfilled its purpose.
> If the above method findUser is called from a Hystrix command I'd expect the command to be marked as successful.

That still complies with what a `HystrixBadRequestException` was intended for. This may be straddling the line of semantics, but a 404 is a result of the client providing bad input, thus a bad request. Thats why I'm pushing for using the existing exception type, and introducing a toggle to allow these requests to count for success metrics.

https://github.com/Netflix/Hystrix/blob/7f5a0afc23aa5ff82320560a04d4c81a45efd67c/hystrix-core/src/main/java/com/netflix/hystrix/exception/HystrixBadRequestException.java#L20-L27

matelang:
It is indeed semantics.

At the end of the day I'm happy to go with any solution that fulfills the end goal here.
Were do you think the toggle would be most suitable in? I'm happy to modify the PR accordingly.

Cheers.

matelang:
Hi @mattnelson.

I've changed the code according to your requests, by introducing `metricsTransient` field on the BadRequestException and treating it accordingly in AbstractCommand.

I hope it's okay now.

arnoldtempfli:
Any update on this?

matelang:
Hi @mattnelson ,

do you want me to do anything else on this PR?
If not can you please merge it?

hystrix-core/src/main/java/com/netflix/hystrix/AbstractCommand.java
```diff
@@ -602,15 +602,23 @@ public void call() {
...
+
                 if (e instanceof RejectedExecutionException) {
+                    circuitBreaker.markNonSuccess();
                     return handleThreadPoolRejectionViaFallback(e);
                 } else if (t instanceof HystrixTimeoutException) {
+                    circuitBreaker.markNonSuccess();
                     return handleTimeoutViaFallback();
                 } else if (t instanceof HystrixBadRequestException) {
-                    return handleBadRequestByEmittingError(e);
+                    HystrixBadRequestException hbre = (HystrixBadRequestException) e;
```
mattnelson: Should be `t` instead of `e` since the else block does an instance of check on `t`.

hystrix-core/src/main/java/com/netflix/hystrix/AbstractCommand.java
```diff
@@ -602,15 +602,23 @@ public void call() {
...
+
+                    if (!hbre.isMetricsTransient()){
+                        circuitBreaker.markSuccess();
+                    }
+
+                    return handleBadRequestByEmittingError(hbre);
```
mattnelson: Shouldn't this be `e` to remain passive?

matelang:
Hi @mattnelson ,

Thanks for your guidance in implementing this.
I've addressed your comments.

Do you know who could I contact for having this merged in?

Cheers,

matelang:
Can somebody please take a look into this? :cry: 
This is causing us issues in production.

matelang:
Can somebody please take a look? :( 
Still causing production issues. I'd really like to avoid building our own Hystrix :cry:

takZhang:
I've never used .  so sorry

gvespucci:
Hi @matelang did you receive any feedback on this? Looks like you gave up at the end... 🥺

matelang:
@gvespucci it seems that nobody is taking care of this repo anymore.

PR #1840, opened 2018-08-15
Fix hystrix.command.default.metrics.rollingPercentile.enabled=false not working

NZvereff:
Made HystrixCommandMetrics abstract and added two different implementations. The implementation choice is based on hystrix.command.default.metrics.rollingPercentile.enabled property


NZvereff:
By the way, RollingCommandEventCounterStreamTest.testShortCircuited() is inconsistent and fails in master as well.

pparth:
Anything new about that?

Issue #1806, opened 2018-05-16
Race condition in HystrixCircuitBreaker in version 1.5.13


tjuszczyk:
In Hystrix 1.5.13 there's a race condition, which makes it possible that HystrixCircuitBreaker object will end up in inconsistent state: ```status==OPEN``` and ```circuitOpened==-1```. This causes ```isOpen()``` method to answer false and circuit will never be closed again, because of:
```
if (status.compareAndSet(Status.CLOSED, Status.OPEN)) {
  circuitOpened.set(System.currentTimeMillis());
}
``` 

Race condition is happening between:
```
 @Override
public void markSuccess() {      
  if (status.compareAndSet(Status.HALF_OPEN, Status.CLOSED)) {
    //This thread wins the race to close the circuit - it resets the stream to start it over from 0
    metrics.resetStream();
    Subscription previousSubscription = activeSubscription.get();
    if (previousSubscription != null) {
      previousSubscription.unsubscribe();
    }
    Subscription newSubscription = subscribeToStream();
    activeSubscription.set(newSubscription);
    circuitOpened.set(-1L);
  }
}
```

and ```onNext``` in HealthCountStream subscriber:
```
@Override
public void onNext(HealthCounts hc) {
// check if we are past the statisticalWindowVolumeThreshold
   if (hc.getTotalRequests() < properties.circuitBreakerRequestVolumeThreshold().get()) {
    // we are not past the minimum volume threshold for the stat window,
    // so no change to circuit status.
    // if it was CLOSED, it stays CLOSED
    // if it was open, we need to wait for sleep window to elapse
  } else {
      if (hc.getErrorPercentage() < properties.circuitBreakerErrorThresholdPercentage().get()) {
      //we are not past the minimum error threshold for the stat window,
      // so no change to circuit status.
      // if it was CLOSED, it stays CLOSED
      // if it was half-open, we need to wait for a successful command execution
      // if it was open, we need to wait for sleep window to elapse
    } else {
       // our failure rate is too high, we need to set the state to OPEN
        if (status.compareAndSet(Status.CLOSED, Status.OPEN)) {
            circuitOpened.set(System.currentTimeMillis());
        }
     }
  }
}
```
Please find the test below, which reproduce the issue:

```
public class HystrixCircuitBreakerRaceConditionTest {

    private static final org.slf4j.Logger LOG = LoggerFactory.getLogger(HystrixCircuitBreakerRaceConditionTest.class);

    private String groupKey;
[...]

Issue #1723, opened 2017-11-30
Circuit Breaker stays open for HystrixObservableCommand and fallback disabled


dstoklosa:
Version affected: 1.5.13
Command: ObservableCommand and fallback disabled

If `ObservableCommand ` is used and circuit gets flipped into OPEN state it will stay permanently in that state if command fallback is disabled. 

`
 @Override
        public boolean attemptExecution() {
            if (properties.circuitBreakerForceOpen().get()) {
                return false;
            }
            if (properties.circuitBreakerForceClosed().get()) {
                return true;
            }
            if (circuitOpened.get() == -1) {
                return true;
            } else {
                if (isAfterSleepWindow()) {
                    //only the first request after sleep window should execute
                    //if the executing command succeeds, the status will transition to CLOSED
                    //if the executing command fails, the status will transition to OPEN
                    //if the executing command gets unsubscribed, the status will transition to OPEN
                    if (status.compareAndSet(Status.OPEN, Status.HALF_OPEN)) {
                        return true;
                    } else {
                        return false;
                    }
                } else {
                    return false;
                }
            }
        }
    }
`

In that case the condition: `if (status.compareAndSet(Status.OPEN, Status.HALF_OPEN))` never returns true.
So hystrix is permanently in half open state and does not allow a new probe request to pass through.

If fallback method is enabled, the circuit will trip back to CLOSED state.

dstoklosa:
This issue can be related to [issue-1721](https://github.com/Netflix/Hystrix/issues/1721 )

bedrin:
Regression in [1.5.13](https://github.com/Netflix/Hystrix/releases/tag/v1.5.13) caused by https://github.com/Netflix/Hystrix/commit/a8203446b90333f005c88cacf91a5dc8faf07c1c

PR #1863, opened 2018-10-05
Use state classes in CircuitBreaker

mwhipple:
Some concerns I have with the changes in: https://github.com/Netflix/Hystrix/commit/c9241ac1a432a3c4eaef3d83a76bf2ccf73c9342#diff-6c58711cb060a8c1b1435ff1e698bb2d

* The specific responsibilities of `circuitOpened` and the current enum value are unclear in terms of how the logic interacts with them.
* `circuitOpened` and the enum value are interdependent but are updated separately; the lack of internal insurance of consistency seems fragile.
* The system can be left in a half-open state if for some reason the single request allowed does not return to one of the mark[Non]Success methods.

There have been several bugs filed related to these concerns. This PR is my thoughts to address the first two and to make the last one more obvious, I'll add a second commit which aims to avoid being stuck in half-open without relying on any external code to do the right thing.

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/HystrixCircuitBreaker.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/HystrixCircuitBreaker.java
@@ -215,10 +215,13 @@ public interface HystrixCircuitBreaker {
 
         // The state during which a single request is expected to be in-flight
         // and will return to call either #markSuccess or #markNonSuccess.
+        // Allows a single request each sleepWindow, though the first request
+        // is always expected to call one of those methods and triggger a state change.
         class HalfOpenState extends State {
+            final long lastAllowed = System.currentTimeMillis();
 
             @Override
-            public boolean allowRequest() { return false; }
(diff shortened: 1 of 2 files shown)


mwhipple:
Resolves #1806

mwhipple:
Resolves #1723

mwhipple:
Subsumes #1640

gvespucci:
Hi @mwhipple , I'm not sure to understand why you closed this one. Asking because I maybe found an issue #1723 related. Thanks

mwhipple:
@gvespucci I closed this just because Hystrix seems abandoned in favor of newer alternatives. Last I knew the issues likely still exist but it seemed unlikely that this PR would be merged so I closed it at some point while doing some GitHub housecleaning.

I can reopen if desired. The code here can be packaged into a usable jar. I think a fair amount of the issues were also introduced in the commit linked to above so using a release prior to that commit also avoids the issues.

gvespucci:
Thank you for the explanation @mwhipple ! I'll have a look a that commit as well. 👋

Issue #1874, opened 2018-10-27
Changed ThreadPoolExecutor setCorePoolSize implementation on Java  10+


gergoistvannagy:
On OpenJdk 11 the implementation of ThreadPoolExecutor's setCorePoolSize was changed and contains an additional constraints comparing the maxPoolSize and corePoolSize values.

If maxPoolSize is not changed but corePoolSize is given as a higher number of default maxPoolSize it throws IllegalArgumentException.

This causes issue here:
https://github.com/Netflix/Hystrix/blob/master/hystrix-core/src/main/java/com/netflix/hystrix/HystrixThreadPool.java#L230-L231

andystanton:
Looks like it's Java 9+.

From the [Java 8 API docs for ThreadPoolExecutor#setCorePoolSize](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html#setCorePoolSize-int-):
> **Throws:**
> `IllegalArgumentException` - if `corePoolSize < 0`

From the [Java 9 API docs for ThreadPoolExecutor#setCorePoolSize](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/ThreadPoolExecutor.html#setCorePoolSize-int-):
> **Throws:**
> `IllegalArgumentException` - if `corePoolSize < 0` or `corePoolSize` is greater than the maximum pool size

Here's a minimal example - it can be triggered if the core size is changed through Archaius:

```java
@Test
void simplestTest() throws Exception {
    var command = new HystrixCommand<>(HystrixCommand.Setter
            .withGroupKey(HystrixCommandGroupKey.Factory.asKey("Group"))
            .andCommandKey(HystrixCommandKey.Factory.asKey("Command"))) {
        @Override
        protected String run() throws Exception {
            return null;
        }
    };
    // ConfigurationManager is from Archaius
    ConfigurationManager.getConfigInstance().setProperty("hystrix.threadpool.Group.coreSize", 11);
    command.execute();
}
```

This exception is thrown when the command is executed:
```
com.netflix.hystrix.exception.HystrixRuntimeException: Command failed while executing.
...
Caused by: java.lang.IllegalArgumentException
	at java.base/java.util.concurrent.ThreadPoolExecutor.setCorePoolSize(ThreadPoolExecutor.java:1535)
	at com.netflix.hystrix.HystrixThreadPool$HystrixThreadPoolDefault.touchConfig(HystrixThreadPool.java:230)
[...]

PR #1877, opened 2018-10-31
Changed ThreadPoolExecutor setCorePoolSize implementation on Java 10+ #1874

TimurD:
Call setMaximumPoolSize before setCorePoolSize to avoid a IllegalArgumentException exception.


andystanton:
Looks like this solves #1874

sachinvd:
I just observed that if you are doing other way round then it would fail. 
Eg.
This solution would work for increasing Thread pool size but if want to decrease thread pool size then it fails.

TimurD:
> I just observed that if you are doing other way round then it would fail.
> Eg.
> This solution would work for increasing Thread pool size but if want to decrease thread pool size then it fails.

Yeah, you are right. I have fixed it.

hystrix-core/src/main/java/com/netflix/hystrix/HystrixThreadPool.java
```diff
@@ -227,8 +227,13 @@ private void touchConfig() {
...
                 }
-                threadPool.setCorePoolSize(dynamicCoreSize);
-                threadPool.setMaximumPoolSize(dynamicMaximumSize);
+                if (dynamicMaximumSize > dynamicCoreSize) {
+                    threadPool.setMaximumPoolSize(dynamicMaximumSize);
+                    threadPool.setCorePoolSize(dynamicCoreSize);
+                } else {
+                    threadPool.setMaximumPoolSize(dynamicMaximumSize);
+                    threadPool.setCorePoolSize(dynamicCoreSize);
+                }
```
benhubert: What's the difference between the `if` and the `else` part? Looks same to me. So if this really works in both cases, I'd change it to ...
```
threadPool.setMaximumPoolSize(dynamicMaximumSize);
threadPool.setCorePoolSize(dynamicCoreSize);
```
:slightly_smiling_face:

umangsingh123:
Do we have any update on this issue . We cant move out of Hystrix right away and we are on Java 11.

TimurD:
Now I also set maximumPoolSize as a corePoolSize setting if corePoolSize is greater to avoid the exception.

TimurD:
It looks like some problems with the project's licenses, but it is not related to my commit and I assume that it is fixed already in the latest version of the project.

hystrix-core/src/main/java/com/netflix/hystrix/HystrixThreadPool.java
tweimer: For the secons statement: Yes.
The second statement is different (dynamicCoreSize vs. dynamicMaximumSize).
Could be written with ternary operator.

PR #1921, opened 2019-01-14
cache publishers for commands by both command group key + command key

ali-jawad:
This fixes a bug whereby if two Hystrix commands have the same command key, although different group keys, only one of them will be reported.

Declaration: I agree to license my contribution under the terms of the APLv2.


dnsbtchr:
Any plans to get this merged soon?

PR #1962, opened 2019-08-21
fix  HystrixTimer when it  initialize ScheduledThreadPoolExecutor does not set RemoveOnCancelPolicy for true

lovingcodesk:
set follow configurations for some command key:

hystrix.command.fallbackcmd.execution.isolation.thread.timeoutInMilliseconds = 1000000000
hystrix.command.fallbackcmd.execution.isolation.strategy = SEMAPHORE
hystrix.command.fallbackcmd.execution.isolation.semaphore.maxConcurrentRequests = 1000
hystrix.command.fallbackcmd.fallback.enabled = true
hystrix.command.fallbackcmd.circuitBreaker.enabled = true

when it  invoke TimerReference.clear(), it does not remove this thread from the queue in ScheduledThreadPoolExecutor,In the case of a large number of requests for long time under this hystrix command key will cause memory leak ,we encounter this problem at our production environment,it Influences service performance

### DIFF
--- a/hystrix-core/src/main/java/com/netflix/hystrix/util/HystrixTimer.java
+++ b/hystrix-core/src/main/java/com/netflix/hystrix/util/HystrixTimer.java
@@ -170,6 +170,7 @@ public class HystrixTimer {
             }
 
             executor = new ScheduledThreadPoolExecutor(coreSize, threadFactory);
+            this.executor.setRemoveOnCancelPolicy(true);
             initialized = true;
         }
 


lovingcodesk:
`Installing oraclejdk8
163$ export JAVA_HOME=~/oraclejdk8
164$ export PATH="$JAVA_HOME/bin:$PATH"
165$ ~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"
166Ignoring license option: BCL -- using GPLv2+CE by default
167install-jdk.sh 2019-07-17
168Expected feature release number in range of 9 to 14, but got: 8
169The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"" failed and exited with 3 during .`

maybe it is not my code problem ,it is because of openjdk

PR #1976, opened 2019-12-01
Fixed possible memory leak in ScheduledThreadPoolExecutor (HystrixTimer.java)

Jagden:
Found what seems to be a possible memory leak inside HystrixTimer.java caused by **cancel** operations called on tasks inside a ScheduledThreadPoolExecutor.
The fix helped us and solved the leakage.
Tasks that were canceled due to timeouts were not removed from the scheduler queues unless **setRemoveOnCancelPolicy(true)** was called on that scheduler on initialization.


tweimer:
Fix #1998

PR #1985, opened 2020-02-04
Official Gradle Wrapper Validation Action

JLLeitschuh:
See: https://github.com/gradle/wrapper-validation-action


JLLeitschuh:
The version of the Gradle wrapper you are using _may_ not be supported (unsupported range is 3.5 to 4.0) so you may have to update by at least one minor version for support.

PR #2022, opened 2021-12-14
fix thread pool error

wangqianqianjun:
When the thread pool is dynamically expanded, and current thread pool maximum size < dynamicCoreSize , will throw exception and stop the current process.  
So first need expand maximum size , then expand core size.


ThinkingMan007:
好的，我已收到。

wangqianqianjun:
For jdk 1.8 that no problem, but when you upgrade your jdk to 16 , will cause some problems.

ThinkingMan007:
好的，我已收到。

PR #2033, opened 2022-11-12
Fix Exception in HystrixThreadPoolDefault.touchConfig() on Java 11

michaelnelson123:
Java 11 ThreadPoolExecutor.setCorePoolSize() throws an IllegalArgumentException if the new coreSize is larger than the current maximumPoolSize. Similarly, setMaximumPoolSize() throws an exception if the new value is less than the current coreSize.


Mohamed-ETS:
Excellent

ThinkingMan007:
好的，我已收到。

